PR #76, opened 2024-03-17
fix: update ota json URL

riptidewave93:
We need to point to the new repo. This is NOT A BLOCKER for 1.0 release.


jwise (approved):
+1, thanks!

PR #84, opened 2024-03-18
update WiFi firmware for BCM43430 to 7.45.98.125

jwise:
Murata have a newer version of the firmware for Infineon / Cypress / Broadcom BCM43430, which at least one user reports meaningfully improves WiFi stability on his printer.  Override the base firmware's 7.45.96.76 (2020-12-21) with 7.45.98.125 (2022-08-16), and also include a new regulatory matrix bundle.


tehmaze:
This PR has resolved the WiFi instability issues on my X1C purchased in February 2024. Before WiFi would crash and disable the menu option.

bani6809 (approved):
works for me

PR #85, opened 2024-03-18
use brcmfmac from backports-5.15.81-1

jwise:
We build in a new kernel that doesn't have any wireless support at all (otherwise, we crash with the backport!), and instead use the brcmfmac backport from 5.15.81-1, aiming it at the new WiFi firmware from #84.  Even Murata says nobody should be using bcmdhd anymore, and so we do not either!

(draft until I actually test a built x1p on my printer)


jwise:
ready for review, works on my printer

bani6809:
how about https://mirrors.edge.kernel.org/pub/linux/kernel/projects/backports/stable/v5.15.148/ ?

bani6809:
user report that the wifi signal indicator is broken, probably using legacy tool to probe it

https://discord.com/channels/1041781310299062282/1188616216936468542/1219938517397016607

jwise:
added a fake iwconfig to shore things up

bani6809 (approved):
works for me

PR #86, opened 2024-03-18
fix kexec_ui after #74

jwise:
kexec_ui removed a little too much not to have the fallback of a printer_ui that we're accidentally loading (oops), and so #74 results in kexec_ui hanging instead of starting (!!).  (This can be solved by using the pwr+stop emergency override.) We add the appropriate bits back to allow kexec_ui to run.


bani6809 (approved):
works for me

PR #113, opened 2024-03-22
Jon/gpiokeys dds

jphannifan:
Gpiokeys was due for a refactor, so this includes revisions on both the QML end and the Python end. 

- DDS now used to communicate button presses with the UI
- The logic for the button actions has been moved from Python to Javascript (excluding the macro setting)
- Atomic rewrite save function added to interpose.cpp
- Button settings now written to its own json file
- Reset to default settings button added

This is a more modular approach and should make it fairly easy to implement custom actions for the door sensor when we're ready to work on that.

Because the gpiokeys settings are being migrated out of printer.json for this, I also introduced an atomic save function that writes to a temporary file, and if the write is successful it renames the file. This is intended to reduce the chances of these settings being corrupted. If the write fails, the temporary file is not renamed. If there's a better way to handle this, I'm open to suggestions


bbl_screen-patch/patches/printerui/qml/X1Plus.js
```diff
@@ -5,6 +5,7 @@
 .import "./x1plus/Stats.js" as X1PlusStats
 .import "./x1plus/MeshCalcs.js" as X1PlusMeshCalcs
 .import "./x1plus/Binding.js" as X1PlusBinding
+.import "./x1plus/GpioActions.js" as X1PlusGpioKeys
```
jwise, line 8: ideally module name should be same as filename; I wil rename module to GpioActions

bbl_screen-patch/patches/printerui/qml/X1Plus.js
```diff
@@ -45,21 +46,33 @@ X1Plus.BedMeshCalibration = X1PlusBedMeshCalibration;
...
 BedMeshCalibration.X1Plus = X1Plus;
 ShaperCalibration.X1Plus = X1Plus;
+GpioKeys.X1Plus = X1Plus;
 
 var _DdsListener = JSDdsListener.DdsListener;
 var _X1PlusNative = JSX1PlusNative.X1PlusNative;
-
 var DeviceManager = null;
 var PrintManager = null;
+var PrintTask = null;
```
jwise, line 62: I realized that it might be possible to pull these out from the Binding shim?  I should try that some day rather than this Awaken nonsense to pass context around.

bbl_screen-patch/patches/printerui/qml/X1Plus.js
```diff
@@ -119,10 +132,17 @@ X1Plus.formatTime = formatTime;
...
 	console.log("X1Plus.js awakening");
 	X1Plus.DeviceManager = _DeviceManager;
 	X1Plus.PrintManager = _PrintManager;
+	X1Plus.PrintTask = _PrintTask;
 	BedMeshCalibration.awaken();
 	ShaperCalibration.awaken();
+	GpioKeys.awaken();
+	// X1Plus.isIdle = _PrintManager.currentTask.stage < _PrintTask.WORKING;
+	X1Plus.hasSleep = _DeviceManager.power.hasSleep;
+	X1Plus.aboutToSleep= _DeviceManager.power.aboutToSleep;
```
jwise, line 147: Ahhh, ok, so, yeah, this will not work.  This captures the immediate value, but does not capture a reference to the binding.  Basically this will capture whether `hasSleep` / `aboutToSleep` right when we `awaken`, but not at any point in the future.  I think the way is to make functions `X1Plus.hasSleep`, etc.

bbl_screen-patch/patches/printerui/qml/X1Plus.js
```diff
@@ -45,21 +46,33 @@ X1Plus.BedMeshCalibration = X1PlusBedMeshCalibration;
...
 X1Plus.emulating = emulating;
 
+var isIdle = null;
+var hasSleep = null;
+var aboutToSleep = null;
+
+function atomicWriteJson(path, json) {
+	_X1PlusNative.atomicSaveFile(emulating + path, JSON.stringify(json));
+}
+X1Plus.atomicWriteJson = atomicWriteJson;
```
jwise, line 74: probably should be named `atomicSaveJson`, by analogue with `saveJson`

bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
```diff
@@ -48,72 +36,71 @@ Rectangle {
...
+    // function pathDialog(inputtxt,inputtitle){
+    //         dialogStack.push("InputPage.qml", {
+    //                             input_head_text : inputtitle,
+    //                             input_text : inputtxt,
+    //                             max_input_num : 50,
+    //                             isUsePassWord : false,
+    //                             isInputShow : true,
+    //                             isInputting_obj : rect_isInputting_obj,
+    //                             output_obj : inputText});    
+    //     }
```
jwise, line 48: rm dead code

bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
```diff
@@ -48,72 +36,71 @@ Rectangle {
...
+    //             }
+    //             setType = "";
+    //         }
+    //     }
+    // }
+    // Text {
+    //     id: inputText
+    //     visible: false
+    //     text: macroFile
+    // }
```
jwise, line 103: rm dead code

bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
```diff
@@ -366,9 +349,18 @@ Rectangle {
...
+                var defaults = X1Plus.GpioKeys.resetToDefaultActions(); 
+                
+                for (var i = 0; i < buttonGrid.children.length; i++) {
+                    var loader = buttonGrid.children[i];
+                    if (loader.item) { 
+                        loader.item.defaultSelection = X1Plus.GpioKeys.getDefault(loader.item.btn, loader.item.pressType);
+                        loader.item.currentIndex = loader.item.defaultSelection;
+                    }
+                }
+            }
```
jwise, line 363: hm, I would rather that `GpioKeys` had the definitive model of what's going on, and this be just a view onto that model.  so in that world we would just `X1Plus.GpioKeys.resetToDefaultActions()`, and then the `item`s would happily just reset themselves based on a binding?

bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
```diff
@@ -400,44 +392,39 @@ Rectangle {
...
+                    }                  
                 }
             }
-            
+          
             Loader {
                 Layout.fillWidth: true
                 sourceComponent: choiceMenu
-                onLoaded: { item.jsonKey = "cfw_power_short"; item.defaultSelection = "4" /* screen lock */; }
+                onLoaded: { item.btn = "cfw_power"; item.pressType = "shortPress";item.defaultSelection = X1Plus.GpioKeys.getDefault(item.btn,item.pressType); item.placeHolder =  X1Plus.GpioKeys.getActionText(item.btn, item.pressType)}
```
jwise, line 421: hm I definitely would prefer that the defaultSelection and the placeHolder be bindings on the btn and pressType rather than repeating a bunch here

bbl_screen-patch/patches/printerui/qml/x1plus/DDS.js
```diff
@@ -45,3 +46,21 @@ _DdsListener.gotDdsEvent.connect(function(topic, dstr) {
...
+        if (datum["command"] == "push_status") {
+            if (gcodeAction() != datum["print_gcode_action"]) {
+                _setGcodeAction(datum["print_gcode_action"]);
+            }
+        } else if (datum["gpio"] == "action" ){
+            var action_code = parseInt(datum["action_code"]);
+            X1Plus.GpioKeys.handleAction(action_code,datum["param"]);
+        }
+    }
+});
```
jwise, line 66: duplicates most of the above?  (also, match on `command`, not `gpio`?)

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
...
+
+
+var settings_file = "/mnt/sdcard/x1plus/buttons.json"
+
+const ACTION_TOGGLE_SCREENSAVER = "Sleep/wake";
+const ACTION_REBOOT = "Reboot";
+const ACTION_PAUSE_PRINT = "Pause print";
+const ACTION_ABORT_PRINT = "Abort print";
+const ACTION_SET_TEMP = "Set temp";
+const ACTION_RUN_MACRO = "Run macro";
```
jwise, line 16: I like having these as strings, though having them be the *same* strings as in the UI could be confusing.  How about having these be `const ACTION_TOGGLE_SCREENSAVER = "TOGGLE_SCREENSAVER"` or something?

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
...
+var buttonConfigs = {
+    "cfw_power": {
+        "shortPress": { action: 4, parameters: {}, default: 4 }, 
+        "longPress": { action: 0, parameters: {}, default: 0 }, 
+    },
+    "cfw_estop": {
+        "shortPress": { action: 2, parameters: {}, default: 2 },
+        "longPress": { action: 3, parameters: {}, default: 3 }, 
+    }
+};
```
jwise, line 35: can these `val`s turn into the `ACTION_`s above?

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
...
+        }
+    } else {
+        console.log(`[x1p] Action not mapped for ${buttonName}, pressType: ${pressType}.`);
+    }
+    return "";
+}
+
+
+function _loadSettings(noload) {
+    if (noload) return;
```
jwise, line 129: presumably 'noload' can be handled by the caller...?

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Creates a deep copy of the default settings object
+ * @return {Object} - Object after defaults are applied
+ */
+function resetToDefaultActions() {
+    var defaultsApplied = JSON.parse(JSON.stringify(buttonConfigs)); 
+    _setGpio(defaultsApplied); 
+    X1Plus.atomicWriteJson(settings_file, defaultsApplied);
+    console.log("[x1p] gpio button settings restored to defaults.");
+    return defaultsApplied;
```
jwise, line 70: I wonder if, like, we set each button one at a time to its default, then the setter can be responsible for doing the write, so we only have to implement that once.

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
+.pragma library
+.import X1PlusNative 1.0 as JSX1PlusNative
+.import "Binding.js" as Binding
+
+var X1Plus = null;
+var _X1PlusNative = JSX1PlusNative.X1PlusNative;
+
+
+var settings_file = "/mnt/sdcard/x1plus/buttons.json"
```
jwise, line 9: ought this be per-printer? everything else in the settings is per-printer (i.e., stored in /config)

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
...
+ * @return {int} - The default setting for the specified button event
+ */
+function getDefault(btn, pressType) {
+    if (buttonConfigs[btn] && buttonConfigs[btn][pressType]) {
+        return buttonConfigs[btn][pressType].default;
+    } else {
+        console.log(`Configuration for ${btn} ${pressType} not found.`);
+        return "";
+    }
+}
```
jwise, line 58: presumably this is an implementation detail, if the default reset in fact binds everything out such that bindings get updated hopefully nobody should ever have to `getDefault` later?

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
...
+
+
+
+/**
+ * Retrieves the action text for a given button type and press type.
+ * @param {string} buttonName - The name of the button ("cfw_power" or "cfw_estop").
+ * @param {string} pressType - The type of press ("shortPress" or "longPress").
+ * @return {string} - The action text.
+ */
```
jwise, line 106: hm could we just have a caller read this straight out of the gpio binding?

images/cfw/opt/gpiokeys.py
```diff
@@ -1,50 +1,46 @@
...
+    5: "Run macro",
+}
 ACTION_REBOOT = 0
 ACTION_SET_TEMP = 1
 ACTION_PAUSE_PRINT = 2
 ACTION_ABORT_PRINT = 3
 ACTION_TOGGLE_SCREENSAVER = 4
-ACTION_NOZZLE_IMAGE = 5
-ACTION_MACRO = 6
+ACTION_MACRO = 5
```
jwise, line 28: same, maybe the internal representation should just be strings here

images/cfw/opt/gpiokeys.py
```diff
@@ -53,108 +49,52 @@ def send_gcode(file_path):
...
+            "action_code": action_code,
+            "param": params,
+            "sequence_id": "0"
+        })
+        gpio_dds_publisher(dds_payload)
+    elif action_code == ACTION_MACRO:
+        file_path = params.get("file_path", "")
+        if file_path.endswith(".py"):
+            args = params.get("arguments","")
+            os.system(f"/opt/python/bin/python3 {file_path} {args}")
```
jwise, line 77: perhaps background it

jwise (commented):
OK I took a pass on this.  there are a handful of things I wanna change, but to prove I'm in it not just to tear things down and make work for other people, I'll make these changes tomorrow :) let me know what you think if you have comments on any of them

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
```diff
@@ -0,0 +1,192 @@
...
+        resetToDefaultActions(); 
+    }
+}
+
+/**
+ * handle button action given dds message
+ * @param {string} action_code
+ * @param {string} datum
+ */
+function handleAction(action_code, datum) {
```
jwise, line 145: also I wonder if this should just live in dds, rather than here.  GpioActions is for configuring how the python script will run, and in theory, many user macros might want to trigger QML events, not just key events?

bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
jphannifan: That's a good suggestion. There's are several sections of code throughout this branch that are very much "let's just get this working and be done" themed and this is one of them.

bbl_screen-patch/patches/printerui/qml/x1plus/DDS.js
jphannifan: This should be changed to _setGpioAction() to match the rest of the format, and yes that should be matching "command" to "gpio_action", not "gpio" to "action". That was a mistake I didn't catch. This is the DDS message format I ended up with: 

>"command": "gpio_action",
>"action_code": action_code,
>"param": params,

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
jphannifan: you're right. I was planning on modifying this, but I wasn't sure if it would be better off in /config/ or on the sd card in /x1plus/prints/{serial}/.

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
jphannifan: yes it looks like the save function gets called a few extra times here. That would clean things up

bbl_screen-patch/patches/printerui/qml/x1plus/GpioActions.js
jphannifan: I like that idea, and I'm wondering should we just do all event handling in QML? Unless there's a benefit to running Python macros from gpiokeys.py, that can just be done from QML and a chunk of gpiokeys.py can be simplified?

jphannifan:
> OK I took a pass on this. there are a handful of things I wanna change, but to prove I'm in it not just to tear things down and make work for other people, I'll make these changes tomorrow :) let me know what you think if you have comments on any of them

Thanks for the comments, all useful and great suggestions. If you don't get to all of these tomorrow I can work on it a bit also. I think we discussed this after you'd already left these comments, but basically we decided the following:

- remove the "Set temp" event aka the "play with fire" feature. Replace with "Activate screenlock" or preferably a shorter string that won't cause the text to be scaled to a comically small size in the combobox. The "Sleep/wake" event only toggles the LCD on/off, and this new one would do the same but also provide users a way to activate the screenlock without having to wait 2-15 min intervals.

That only leaves "Run macro" that's still handled in Python, it seems to me that it simplifies things a bit to handle all events in QML.

bani6809 (approved):
printer didn't dead. approved.

bbl_screen-patch/patches/printerui/qml/settings/ScreenLock.qml
```diff
@@ -5,37 +5,40 @@ import QtQml 2.12
...
-            return false;
-        } else {
-            return true;
-        }
+
+    ZRoundedImage {
+        id: lockimg
+        width:1180
+        height:720
+        anchors.fill: parent
```
jwise, line 37: these constraints conflict

bbl_screen-patch/patches/printerui/qml/settings/ScreenLock.qml
```diff
@@ -5,37 +5,40 @@ import QtQml 2.12
...
-            return true;
-        }
+
+    ZRoundedImage {
+        id: lockimg
+        width:1180
+        height:720
+        anchors.fill: parent 
+        originSource: "file://" + lockImage
+        visible: X1Plus.fileExists(lockImage)
```
jwise, line 39: this will not work correctly for the built in qrc:/printerui/image?

bbl_screen-patch/patches/printerui/qml/settings/ScreenLock.qml
```diff
@@ -5,37 +5,40 @@ import QtQml 2.12
...
-            id: lockimg
-            anchors.fill: top
-            cornerRadius: parent.radius
-            originSource: "file://" + lockImage
-            visible: imgExists(lockImage)
+
+    Binding on passcode { //possibly a redundant property
+        value: numberPad.number
+        when: numberPad.target == lockText
     }
```
jwise, line 31: I'm not sure I understand this?  this seems to override the passcode entirely such that any passcode will unlock the printer?

bbl_screen-patch/patches/printerui/qml/settings/ScreenLock.qml
```diff
@@ -131,20 +137,22 @@ Rectangle {
...
-            radius: height / 2
+            radius: parent.height / 2
             color: Colors.gray_700
-                
+        
             Text {
+                id: slideText
                 horizontalAlignment: Text.AlignHCenter
                 verticalAlignment: Text.AlignVCenter
+                anchors.fill: parent
```
jwise, line 150: I will go re-minimize these diffs

Issue #116, opened 2024-03-23
decide if we're going to support cifs/smb3 or remove it


bani6809:
cifs and smb3 are in the kernel.

nodev   cifs
nodev   smb3

we are missing the cifs-utils /sbin/mount.smb3 & /sbin/mount.cifs helpers

if we are going to support cifs/smb3 we should add the helpers, if we are not then we should remove from kernel and get some memory back.

PR #118, opened 2024-03-24
add mount.cifs and mount.smb3 to squashfs

bani6809:
this is for issue #116

adds crc32's build of mount.cifs from cifs-utils

make a file /mnt/.smbcredentials containing:

username=user
password=pass

mkdir /mnt/samba

then do

mount.smb3 //windowspc/fileshare /mnt/samba -o credentials=/mnt/.smbcredentials


jwise (approved):
I think we can afford the 124 kB.

Issue #112, opened 2024-03-21
sighting: printer userspace slows to a crawl (OOM-like behavior)


jwise:
**Describe the bug**
The old memory leak behavior has been sighted again, where sometimes a printer will slow to a crawl during a print, Gcode will be sent a few lines at a time, the UI will be unresponsive, SSH may be unresponsive.

Update this bug with further sightings.

**To Reproduce**
No known repro yet.

**Sightings**
1. https://discord.com/channels/1041781310299062282/1188616216936468542/1220182969373819010

PR #119, opened 2024-03-24
add S99perf_log

bani6809:
debugging tool for end users to try to catch issue #112 

they change /etc/init.d/.x1plus_config and change PERF_LOG=on and reboot (or /etc/init.d/S99perf_log start) and then periodic free and ps are written to /mnt/sdcard/perf_log.txt

### DIFF
--- a/images/cfw/etc/init.d/.x1plus_config
+++ b/images/cfw/etc/init.d/.x1plus_config
@@ -1,1 +1,0 @@
-PERF_LOG=off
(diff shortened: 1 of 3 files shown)


jwise:
this seems like part of a broader printer config question that we have.  I wonder if this should go in, like, /mnt/sdcard/x1plus/printers/{...}/config.json, and then it should log to /mnt/sdcard/x1plus/printers/{...}/logs/perf.log ?

PR #120, opened 2024-03-25
fix the need to LD_PRELOAD libffi into python

### LINKED ISSUE #111: fix the LD_PRELOAD chaos in Python
**Describe the bug**
For some reason, we need to `LD_PRELOAD` libffi into the Python that we build.  I should do some linker hacking or something to make the build of Python that we distribute not have this problem, because this sucks.

**To Reproduce**
1. Try to load `ctypes` into a `python3` that you invoke at the command line without any noise.

jwise:
patchelf python and its libraries to load from /opt/python/lib, and also patchelf ctypes to load libffi, and also strip python libraries to save quite a bit of space (fixes #111)

### DIFF
--- a/images/cfw/etc/init.d/S82gpiokeys
+++ b/images/cfw/etc/init.d/S82gpiokeys
@@ -2,7 +2,6 @@
 
 start() {
 	printf "Grabbing gpiokeys"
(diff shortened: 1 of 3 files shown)


bani6809 (approved):
installed and printed test model and it do not asplode. approved.

PR #125, opened 2024-03-26
cleanup mb counter

bani6809:


jwise (approved):
untested but looks right

Issue #81, opened 2024-03-18
X1Plus startup failed.


dmiart1235123:
**Describe the bug**
When fat32 is not the first partition on sdcard it fails to boot into X1Plus. This is fixed by re-formating sd card to have sd card as first partition. 

**To Reproduce**
1. Have two partitions on sd card(how RPI image partitions it)
2. Install X1Plus 
3. Everything will install without error
4. On first boot into X1Plus will show console error "Mouting base layer 00.00.28.55.squashfs\n  X1PLUS startup failed. initramfs shell is available on UART at 1500000 band."

**Expected behavior**
Expected to boot into X1Plus

**Screenshots/Photos**

![image](https://github.com/X1Plus/X1Plus/assets/155584166/f7dd0adf-dcbd-4525-a2ed-02109b4bf98e)

jwise:
Oh yeah this is definitely a bug.  Probably we should just have the installer cry instead?

psiberfunk:
Def. feels like we should tell people to format with a fresh SD card in such a situation and bomb out.

dmiart1235123:
Yeah, I would add a warning somewhere to tell people to re-format cards. I am sure there will be many, many people who just reuse sd cards from other projects. Mine was previously used in RPI that's why it has two partitions. The crappy part, not all windows versions play nicely with two or more partitions on sd card. So some people might not even realize they have two partitions on sd card. Windows 10 finally has a proper support for that though. It will not detect ext filesystem though.

psiberfunk:
> Yeah, I would add a warning somewhere to tell people to re-format cards. I am sure there will be many, many people who just reuse sd cards from other projects. Mine was previously used in RPI that's why it has two partitions. The crappy part, not all windows versions play nicely with two or more partitions on sd card. So some people might not even realize they have two partitions on sd card.(windows might not assign a letter to other partitions) Windows 10 finally has a proper support for that though. However, It will still not detect ext filesystem.

WSL2 actually does, but that's besides the point.. a proper partition nuking should do it .

jomixlaf:
I believe I was the first person (documented) running thru this issue yesterday. In my case it was the SDCard already in the Bambu for recording. When it finished install the CFW it failed with this error, I then format the SD card via the Bambu itself, still an issue. I will try to formet it with Mac OS once I'm at home.

CobraXLazer:
Same problem here, my stoke went away!

jwise:
Your stoke will soon return.

[...]

PR #130, opened 2024-03-28
check that sd is formatted properly

bani6809:
addresses issue #81

### DIFF
--- a/installer/install.py
+++ b/installer/install.py
@@ -214,9 +214,9 @@ def validate_sd():
         dev, pt, fs = m.split(' ')[:3]
         if pt == "/mnt/sdcard":
             if dev != "/dev/mmcblk2p1":
-                report_failure(f"The SD card has more than one partition and is improperly formatted.<br><br>Please format the SD card correctly and restart the installer.")
+                report_failure(f"The SD card has more than one partition and is improperly formatted. Please format the SD card correctly and restart the installer.")
             if fs != "vfat":
-                report_failure(f"The SD card is not formatted as vfat.<br><br>Please format the SD card correctly and restart the installer.")
+                report_failure(f"The SD card is not formatted as vfat. Please format the SD card correctly and restart the installer.")
             found_sdcard = True
     if not found_sdcard:
         report_failure(f"No SD card found. Please insert a vfat formatted SD card of at least 16gb and restart the installer.")
--- a/installer/install.py
+++ b/installer/install.py
@@ -31,6 +31,7 @@ import time
 import filecmp
 import glob
 import subprocess
+import re
 
 # XXX: have config for filesystem size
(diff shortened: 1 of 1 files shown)


installer/install.py
```diff
@@ -205,7 +206,27 @@ def traverse_ino(path, ino):
...
+    except subprocess.CalledProcessError:
+        report_failure("is_vfat blkid {device} failed.")
+
+    partition_type = blkid_output.decode('ascii').strip()
+    print( partition_type )
+    return re.search(r"vfat|FAT32|FAT16", partition_type) is not None
+
+#def format_sdcard():
+
+if os.path.exists("/dev/mmcblk2p2"):
```
jwise, line 221: I think this actually will not solve the problem where the SD Card is formatted with an extended partition table -- in that case, we get mmcblk2p1 and mmcblk2p5 (!).

I wonder if the thing we actually mean to do is something like:
```python
with open("/proc/mounts", "r") as f:
    mounts = f.read().strip().split('\n') 
found_sdcard = False
for m in mounts:
    dev, pt, fs = m.split(' ')[:3]
    if pt == "/mnt/sdcard":
        if dev != "/dev/mmcblk2p1":
            fail...
        if fs != "vfat":
            fail...
        found_sdcard = True
if not found_sdcard:
    fail...
```

... which tests for precisely what we mean: "the SD card that we are about to write to is mounted in precisely the way that we will mount it on the way up when we boot X1Plus"?

installer/install.py
```diff
@@ -205,7 +206,22 @@ def traverse_ino(path, ino):
...
+    for m in mounts:
+        dev, pt, fs = m.split(' ')[:3]
+        if pt == "/mnt/sdcard":
+            if dev != "/dev/mmcblk2p1":
+                report_failure(f"The SD card has more than one partition and is improperly formatted. Please format the SD card correctly and restart the installer.")
+            if fs != "vfat":
+                report_failure(f"The SD card is not formatted as vfat. Please format the SD card correctly and restart the installer.")
+            found_sdcard = True
+    if not found_sdcard:
+        report_failure(f"No SD card found. Please insert a vfat formatted SD card of at least 16gb and restart the installer.")
```
jwise, line 222:
```suggestion
                report_failure(f"The SD card's partition scheme is unsupported. Format the SD card from the on-printer menu, then restart the installer.")
            if fs != "vfat":
                report_failure(f"The SD card's filesystem is unsupported. Format the SD card from the on-printer menu, then restart the installer.")
            found_sdcard = True
    if not found_sdcard:
        report_failure(f"SD card was not mounted in the expected location.  If the SD card is inserted, please file a bug.")
```

jwise (approved):
+1 assuming you tested it, I have suggested wording changes

PR #135, opened 2024-03-30
more correct checks

bani6809:


jwise (approved):
+1, I'm not entirely convinced that it changes a whole lot but it can potentially avoid users getting into weird configurations where they don't get to use their entire SD card

PR #136, opened 2024-03-30
update many strings to use qsTr

jwise:
Many strings in the X1Plus UI were not translatable, or otherwise break original translations.  Revert the ones that break original translations, and wrap everything else in `qsTr()` (and use `qsTr().arg()` where appropriate).


jphannifan (approved):
отпускает!

jphannifan:
kexec_ui: fully translatable

printer_ui: "Pull to unlock" for lockscreen doesn't translate, HardwarePage dropdown display text doesn't translate

everything else looks good.

PR #138, opened 2024-03-31
add ignore option

bani6809:


bbl_screen-patch/patches/printerui/qml/x1plus/GpioKeys.js
```diff
@@ -34,6 +36,7 @@ const BUTTON_MAPPING_OLD = {
     "4": ACTION_TOGGLE_SCREENSAVER,
     /* "5": nozzle cam no longer exists */
     /* "6": run macro no longer exists */
+    "7": ACTION_NONE,
```
jwise, line 39: This is just for migration from the old settings structure. Since this didn’t exist in the old settings structure (and definitely not with the value “7”!) we do not need to add it here.

jphannifan (approved):

jwise (approved):
+1

PR #141, opened 2024-03-31
Jwise/fix printingpng

jphannifan:
This fixes the image display issue and cleans up variable names. I have tested and confirmed it's fixed on my device, and anyone who wants to test it can do so by editing printer.json like so:

`"cfw_print_image": "/mnt/sdcard/test1.png",`
`"cfw_home_image": "/mnt/sdcard/test2.png",`

replace the filepaths with your own, of course. 

I think we should remove the hardcoded paths, because it only seems to be a source of confusion for users trying to set custom background images.


jwise (approved):
OK with me.  Though I think that we should store stuff in sdcard/x1plus/printers/blah/settings.json for this one, not printer.json.  Getting people to scp files over at all is hard enough, let alone having them edit an on-printer json that could result in making bbl_screen really angry.

Bonus points: if a settings file is corrupted maybe we should emit a x1plus HMS error.

jwise:
please use `git rebase` (or `git pull --rebase`), not `git merge`.  rebased and pushed :)

Issue #131, opened 2024-03-28
[bug report] Bed rammed into nozzle


tscharly:
Few days ago i had to pause a print job because of a blob of filament around the nozzle which i had to remove.
Manually moved the bed down and the printhead to the backside.
After resuming printjob, the bed was hammered into the nozzle at full speed and force at two different positions.

**Additional Details:**
 - Printer type: X1C
 - X1Plus version: 1.0

tscharly:
![IMG_20240326_164645_502.jpg](https://github.com/X1Plus/X1Plus/assets/139580028/e1a5705d-d155-49fc-9c31-9abb4a4e1940)

![IMG_20240326_164638_803.jpg](https://github.com/X1Plus/X1Plus/assets/139580028/16cb6294-b689-4c7d-844c-33c7afc4294a)

RobinWhitfield:
I have a suspicion that this is an OEM bug.

When you say you moved the bed down to clear the blob of filament, did you move it back up *at all* after clearing it?

Can you remember *exactly* what movements you requested from the printer before and after clearing the clog?

Moving the bed really should not be possible at all when you've paused a print as there's no way for the printer to home Z again without potentially colliding with a print. This option should be disabled.

scooterthedev:
Funny enough... I just had the same thing happen to me today, but not while printing. I went to the console, and clicked on lower bed, and the bed flew up, and with the nozzle with so much power, I actually jumped lol. Now there is a huge hole, like the same size as in the photo above, in my bed :(.

RobinWhitfield:
@scooterthedev, can you clarify exactly what you did and what you were doing immediately beforehand?

Your comment mentions the console, but I'm guessing you don't mean the G-code/SSH console built into X1Plus.

Which x1plus version are you running?

jphannifan:
> Funny enough... I just had the same thing happen to me today, but not while printing. I went to the console, and clicked on lower bed, and the bed flew up, and with the nozzle with so much power, I actually jumped lol. Now there is a huge hole, like the same size as in the photo above, in my bed :(.

Which button did you press? 

Some extra precaution must be taken when using the console. 

From your description, it sounds like you entered the gcode command generated from pressing the button on the console screen called "Bed Down" or "Bed Up"? If left unedited, this sends the command
`G0 Z10 F1200`

There are no hard endstops on this printer, so I would advise using the lowest acceleration values and smallest increments for movement when testing any gcode commands for motion control.

[...]

PR #143, opened 2024-04-01
Console: enforce relative movement before bed movement command!

jwise:
Hopefully fixes @scooterthedev 's sighting of #131, which is probably a different syndrome than the OP of #131.


bani6809 (approved):
appears to work.

PR #147, opened 2024-04-01
feat: bring back initial ota script

riptidewave93:
This gives us an initial pass at an OTA engine that can be used via shell, and called via bbl_screen with the get_info arg.

Idea is this is a first pass, and can later move it to our shared x1plus python daemon for interfacing with via mqtt. When we do that, this script will be rewritten to just make mqtt calls to said service.

However, for now this gets us something out for advanced users who can SSH, and gives us an initial building block.


riptidewave93:
Killing this, since we are working on a proper UI solution, and not just a "workaround" tool.

PR #155, opened 2024-04-05
interpose: do not attempt to inject languages into something that does not look like a map

### LINKED ISSUE #146: translation does not work correctly on FW < 01.07.02.00
kexec_ui crashes on 01.07.00.00 now.

We should probably just detect bblap version and refuse to do the language injection thing in kexec_ui on any version older than FW R / the officially supported bblap version.  It is pretty brittle and we can't really maintain it for all of bbl_screen history, I suppose.  Maybe we could get BBL to add a hook for us for this...

jwise:
On old firmwares -- like 00.00.24.13 -- the language injection process will try to peek through data structures in a way that is not valid for that version of bbl_screen's memory layout.  Now, we make sure that there is at least *some* kind of QMap where we expect there to be one when we're going to inject languages.  This means that kexec_ui on 00.00.24.13 will not be translated correctly, but at least we won't crash.

Fixes #146.


jphannifan:
I'm testing this one now. Will report back soon

PR #159, opened 2024-04-07
syslog_shim encoding fix

jphannifan:
Prevent syslog_shim.py from being killed by UnicodeDecodeErrors

If any non UTF-8 characters end up in syslog, syslog_shim.py throws a UnicodeDecodeError and does not run. Opening the file with errors="replace" handles this error.


jwise (approved):
+1, good catch

PR #160, opened 2024-04-07
Feed frames to libvncserver from the framebuffer without restarting bbl_screen.

### LINKED ISSUE #173: Simultaneous VNC + FB
It would be nice to not have to do SSH shenanigans to start VNC, and also to be able to go back to using the printer after VNCing seamlessly.

jwise:
This is a good one!  We interpose calls to libdrm that would set up the framebuffer, and when a framebuffer is setup, we run a VNC server based on that.  To get performance up a little bit, we interpose calls to `QRegion` to make sure that we only update the parts of the screen that we need to.  To rotate the framebuffer and convert the pixel format, we use NEON intrinsics.  Performance is somewhat better than the native VNC server, though the printer display does lag a little bit when VNC is connected.  The lag goes away when VNC is disconnected.

I considered writing a Qt Platform Plugin.  Doing so would be outright chaos, especially with the internal versioning changes of Qt -- we would need exact headers to make that work.  So I did it the more fun way!

To test this, try:

`/opt/x1plus-test-settings.py vnc.enabled=true`

You should be able to VNC to your printer using your LAN code as a password.

`/opt/x1plus-test-settings.py vnc.password=lol`

Instead of your LAN code, you can VNC using the password "lol".

`/opt/x1plus-test-settings.py vnc.password=""`

You can VNC with no password.

`/opt/x1plus-test-settings.py vnc.password=None`

You once again will use your LAN code as a password.

`/opt/x1plus-test-settings.py vnc.enabled=None`

VNC will be turned off.

Fixes #173.


jwise:
the situation is much better now.

riptidewave93 (approved):

bani6809:
build is busted. bbl_screen doesn't start up.

bani6809 (approved):
very performant. appears to work as advertised. 👍

jphannifan:
It's so speedy! nice

jphannifan (approved):

PR #161, opened 2024-04-07
fix: don't prompt for updates on missing hardware

riptidewave93:
In cases where an AP board is ran outside of a printer without an MC/TH board, X1Plus always errors for firmware version for these boards. This fix disables the update prompt for devices that do not report an SN, aka, they are not attached to the printer.

We do this by parsing m.sn which is "N/A" for missing core devices, but empty for optional things (ams).


riptidewave93:
Note that I need to do more testing before merging, and any assistance in testing this doesn't break existing functionality would be appreciated. :)

riptidewave93:
Finished my testing, this works as expected from what I can see.

jwise (approved):
LGTM

PR #179, opened 2024-04-11
ConsolePage: fix wrapping

### LINKED ISSUE #145: console text does not wrap properly after translation fixes
image

jwise:
When we added translation support to ConsolePage, we got rid of all of the manual newlines in the placeholder text.  Unfortunately, placeholder text is not subject to the `wrapMode` policy.  We simulate placeholder text with real text that is gated on `outputText.length` -- and then modify the rest of the properties to match the placeholder property.  (Since the placeholder text is now plaintext, not rich text, we replace the `<br>`s with `\n`s.)

Fixes #145.


jphannifan (approved):
I'm thinking we should ditch the `TextArea` type on ConsolePage.qml and use a ListView. Then we also no longer have to manually format the placeholder text here to make sure it wraps properly. But on top of that, it would also allow us to adjust the color of the text, like so:
image

Easier said than done, but it's a reasonable goal for 1.2 or 1.3 I think

jwise:
Yeah, I agree that this should eventually be a ListView.

PR #180, opened 2024-04-11
Bump translations before 1.1

### LINKED ISSUE #175: 1.1 localization updates
make sure that we grab the latest localization stuff for 1.1 before we release it

jwise:
Fixes #175.

Also fixes a silly error in translation upload that would create bad translation builds later.


jphannifan (approved):
good catch!

PR #186, opened 2024-04-16
add df, sleep for 10 instead of 30

bani6809:


riptidewave93 (approved):
Lgtm but maybe we also want uptime? Just my 2c.

jphannifan:
This looks good to me too, but if I had to change anything it'd be moving to python so you can use Python's `logging`.

This would print to a rotated and timestamped log file, and if you wanted you can use logging levels

```
import subprocess
import time
from logger.custom_logger import CustomLogger #/opt/logger/custom_logger.py

def cmd_output(command):
    return subprocess.run(command, capture_output=True, text=True).stdout

log_dir = "/mnt/sdcard/perf.log"
perf_log = CustomLogger("perf", log_dir ,500000,1,True)  

perf_log.info('Perf log started')

while True:
    perf_log.info('Disk Free: %s', cmd_output(['df']).strip())
    perf_log.info('Memory Usage: %s', cmd_output(['free']).strip())
    perf_log.info('Process Snapshot: %s', cmd_output(['ps', '-auwxf']).strip())
    time.sleep(10)
```

jphannifan:
And device uptime as per @riptidewave93 's suggestion, which I stole from /usr/bin/sys_monitor.sh
```
echo `cat /proc/uptime | awk '{print $1}' | awk -F. '{print "up_sec:"$1}'`
```

jphannifan (approved):

PR #195, opened 2024-04-22
Upgrade base firmware to 01.07.03.05.

jwise:
This changeset upgrades us to the beta firmware version, 01.07.03.05 (aka 00.00.32.17).  To do this, we:

* Modify interpose.cpp to support a new DdsNode version.  (It also supports the old DdsNode version, for kexec reasons.)
* Ports QML patches forward to the new version.
* Lays the groundwork for interposing dbus-c++: connect DBus-C++ to the session bus, rather than the system bus, in emulation.


bani6809 (approved):
it do not asplode. approved

PR #196, opened 2024-04-22
Bind the new BDbus library to QML

jwise:
Bambu Lab are moving to DBus now in the newest beta firmware, and all things told, this is a lot better than DDS.  In this PR, we introduce a mechanism to plumb DBus function calls and signals into QML, as well as outbound DBus function calls, so we can take action inside the UI.  Right now we implement only a `ping` command and a `log` signal, which you can try with:

```
dbus-send --session --type=signal / x1plus.screen.log string:'{"text": "hello, world"}'
dbus-send --session --print-reply --dest=bbl.service.screen /bbl/service/screen bbl.screen.x1plus.ping string:'{"text": "hello from DBus"}'
```

These commands are in emulation; on the real system, you will want to use `--system` instead of `--session`.  This lays the groundwork for being able to call RPCs into QML, and report events into QML.

Additionally, you can run an "X1Plus hello world" service elsewhere on the system which responds to RPCs from inside QML.  Try running the following elsewhere on your host:

```python
from dasbus.connection import SessionMessageBus
from dasbus.server.interface import dbus_interface
from dasbus.typing import Str
import json

@dbus_interface("x1plus.hello")
class X1PlusHello(object):

    def PingPong(self, arg: Str) -> Str:
        print(f"pingd: {arg}")
        arg = json.loads(arg)
        return json.dumps(f"Pong {arg}")

print(X1PlusHello.__dbus_xml__)

bus = SessionMessageBus()
bus.publish_object("/x1plus/hello", X1PlusHello())
bus.register_service("x1plus.hello")

from dasbus.loop import EventLoop
loop = EventLoop()
loop.run()
```

Then, you can trigger bbl_screen to call into this:
```
$ dbus-send --session --print-reply --dest=bbl.service.screen /bbl/service/screen bbl.screen.x1plus.TryRpc string:'{"text": "hello from DBus"}'
method return time=1714115957.022830 sender=:1.2523 -> destination=:1.2525 serial=11 reply_serial=2
   string "{"text":"hello from DBus","resp":"Pong hello"}"
```

This is two DBus calls for the price of one: bbl_screen listens to DBus, and then it spits out a DBus call, and then returns what it heard back from that DBus call.

### DIFF
--- a/bbl_screen-patch/Makefile
+++ b/bbl_screen-patch/Makefile
@@ -64,8 +64,11 @@ base.ts: printer_ui.so kexec_ui.so
 %.so: %.c
 	$(TARGET_CC) -o $@ $< $(CFLAGS) $(LDFLAGS)
 
-%.so: %.ui.o interpose.cpp interpose.moc.h
+kexec_ui.so: kexec_ui.ui.o interpose.cpp interpose.moc.h
 	$(TARGET_CC) -o $@ $< interpose.cpp workaround_ifaddrs.c $(CXXFLAGS) $(LDFLAGS) -I $(QT5_INCL_PATH)
+
+printer_ui.so: printer_ui.ui.o interpose.cpp interpose.moc.h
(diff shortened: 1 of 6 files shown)


jwise:
Yep.  The other big win will be gpiokeys, I think.

riptidewave93 (approved):

PR #198, opened 2024-04-22
chamber light

jphannifan:
Interpose chamber light. This works identically to the backlight function

Usage:
Turn on:  `X1PlusNative.updateChamberLED(1)`
Turn off: `X1PlusNative.updateChamberLED(0)`

### DIFF
--- a/bbl_screen-patch/interpose.cpp
+++ b/bbl_screen-patch/interpose.cpp
@@ -275,9 +275,8 @@ public:
     }
 
 public:
-    int chamberledSetting = 255;
+    int chamberledSetting = 1;
     Q_INVOKABLE void updateChamberLED(int val) {
-        chamberledSetting = (val == 1) ? 255 : 0;
         std::string valueText = std::to_string(chamberledSetting);
         int fd;
         if ((fd = open("/sys/devices/platform/gpio-leds/leds/sys_led/brightness", O_RDWR)) >= 0) {
--- a/bbl_screen-patch/interpose.cpp
+++ b/bbl_screen-patch/interpose.cpp
@@ -274,6 +274,18 @@ public:
         }
     }
 };
+public:
+    int chamberledSetting = 255;
+    Q_INVOKABLE void updateChamberLED(int val) {
+        chamberledSetting = (val == 1) ? 255 : 0;
+        std::string valueText = std::to_string(chamberledSetting);
+        int fd;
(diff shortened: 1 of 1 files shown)


riptidewave93 (approved):

bbl_screen-patch/interpose.cpp
```diff
@@ -273,6 +273,17 @@ class X1PlusNativeClass : public QObject {
             close(fd);
         }
     }
+
+public:
```
jwise, line 277: this is already `public`

bbl_screen-patch/interpose.cpp
```diff
@@ -273,6 +273,17 @@ class X1PlusNativeClass : public QObject {
...
+    int chamberledSetting = 1;
+    Q_INVOKABLE void updateChamberLED(int val) {
+        std::string valueText = std::to_string(chamberledSetting);
```
jwise, line 280: This does not actually set `chamberledSetting`!!

bbl_screen-patch/interpose.cpp
```diff
@@ -574,6 +597,16 @@ SWIZZLE(ssize_t, write, int fd, const void *p, size_t n)
         } else {
             printf("backlight off\n");
         }
+    } else if (fd == chamber_led_fd) {
+        printf("interposed write() on backlight, ");
+        if (*(char*)p != '0') {
+            printf("writing %d instead\n", native.chamberledSetting);
+            std::string valueText = std::to_string(native.chamberledSetting);
+            next(fd, valueText.c_str(), valueText.length());
+            return n;
```
jwise, line 606: This appears to "AND" the chamber LED with X1Plus.  Do we want to override it both directions?  Or just as an AND?

jwise:
Hm, I don't understand the intent here.  How does this differ from `DeviceManager.isLightOn`?  That is what the "chamber light" button in the GUI does.  A quick `grep` also indicates that there are many parts of the system that touch the chamber light, not just `bbl_screen`.

Beyond that, what is the intended use of this?

bbl_screen-patch/interpose.cpp
jphannifan: you're right, `chamberledSetting` was set incorrectly in a now deleted line. I'm not going to fix this one, because this PR does not need to exist

jphannifan: you're right, `chamberledSetting` was set incorrectly in a now deleted line. I'm not going to fix this one, because this PR does not need to exist

PR #199, opened 2024-04-22
gcodegenerator.js revisions

jphannifan:
GcodeGenerator.js - The aim here was to correct several incorrectly defined or formatted gcode commands and organize the code a bit. The long term goal is to make it easy for us to add and call new macros. 
- Several incorrectly defined or formatted gcode commands were corrected.
- Commands added: M1005, G92, M211, G2, M205, M220, M201, M302, M412, G29.2 
- Added interpolation function for print speed. Trend fitting has been detailed here, and this can be rewritten and used as wiki content in the future: https://github.com/jphannifan/x1plus-testing/blob/main/BL-speed-adjust.md
- 'rampSpeedLevel' function added
- ColdPull macro and chamber preheat macro added

X1Plus.js
- removed X1Plus.GcodeMacros (BedTramming.qml and ShaperCalibration.js can just call GcodeGenerator directly)
- Removed sequence_id: "420" from X1Plus.sendGcode

ConsolePage.qml
- Gcode commands were previously in two lists, and I've moved them into a single object
- Command history feature fixed. It now lives in the "hotkey" bar as the first button in the panel. We might replace this with an icon in the future
- Added warning dialog if gcode or shell command is entered while printer is printing.
- Added M1005 to command hotkeys
- renamed some properties for clarity

BedTramming.qml and ShaperCalibration.js
- Updated to call new Gcode functions

I still feel like ConsolePage.qml could be a lot more useful with some minor adjustments to how gcode is input and entered.


bani6809 (approved):
lightly smoke tested, it do not asplode. approved

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -230,61 +145,73 @@ function M975(enabled = true) {
...
+        speedPercentage = 100;
+    }
+    // Convert percentage to a fraction, use Math.floor() to keep our target % the same as Bambu's reported %
+    var speedFraction = speed_interp.speed_fraction(speedPercentage);
+    
+    // Calculate acceleration magnitude from speed fraction based on log trendline
+    var accelerationMagnitude = speed_interp.acceleration_magnitude(speedFraction);
+    
+    // Interpolate feed rate from R
+    var feedRate = speed_interp.feed_rate(speedPercentage);
```
jwise, line 182: I feel uneasy about this being how things are derived, if this is going to drive the GUI for how speed is calculated.  In this case, '100% speed' *will set the speed to 105%*!  Obviously we must not do that: that will break prints that are carefully calibrated to be at the VFR limit (which seems to be a thing users do...).

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -330,73 +254,299 @@ function G4(sec = 90) /* pause */
...
+        const rampProgress = currentLayer - startingLayer;
+        let speedChange = rampProgress * delta;  
+        let newSpeed = startingSpeed + speedChange;
+        if (startingSpeed < targetSpeed) {
+            return Math.min(newSpeed, targetSpeed);
+        } else {
+            return Math.max(newSpeed, targetSpeed);
+        }
+    }    
+}
```
jwise, line 552: This looks like a straightforward linear interpolator.  Should it be in GcodeGenerator or in whatever timer is actually doing the LERP?

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -55,106 +13,48 @@ function createGcode(command, params = {}) {
...
-    let header = M1002({action_code:13,action:0});
-    let footer = M1002({action_code:0,action:0});
-    if (nozzleTemp>0 && nozzleTemp <= 300){
-        header += M109(nozzleTemp);
-        footer += M109(0);
-    }
-    if (bedTemp>0 && bedTemp <= 300){
-        header += M140(bedTemp,true);
-        footer += zzzM140(0,false);
-    }
```
jwise, line 16: In the old mechanism, we M1002(13) *before* we M140(bedTemp).  This is, at least, a behavioral change in the new one; is it intentional?

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -55,106 +13,48 @@ function createGcode(command, params = {}) {
...
 
-    let gcode = M73(0,3); // update timeline
-    gcode += M201(100); // set Z max accel
-    gcode += G90(); // absolute coords
-    gcode += M400(1); // pause
-    gcode += M17(1.2, 1.2, 0.75); // stepper current Z=0.75
-    gcode += M201(1000); // set Z max accel
-    gcode += G28(HOMING.XYZ); // home
-    gcode += G92_1(); // unknown Bambu shaper code
```
jwise, line 122: This "unknown bambu shaper code" appears to be missing in the new one.

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -55,106 +13,48 @@ function createGcode(command, params = {}) {
...
+    old_retract_length: () => "[old_retract_length]",
+    new_retract_length: () => "[new_retract_length]",
+    old_filament_temp: () => "[old_filament_temp]",
+    new_filament_temp: () => "[new_filament_temp]",
+    filament_type: (idx) => `{filament_type[${idx}]}`,
+    x_after_toolchange: () => "[x_after_toolchange]",
+    y_after_toolchange: () => "[y_after_toolchange]",
+    z_after_toolchange: () => "[z_after_toolchange]",
+    old_filament_e_feedrate: () => "{old_filament_e_feedrate}",
+    new_filemnt_e_feedrate: () => "{new_filemnt_e_feedrate}",
```
jwise, line 37: Is this our typo, or theirs?  If theirs, do we want to homogenize it here?

bbl_screen-patch/patches/printerui/qml/printer/BedTramming.qml
```diff
@@ -125,7 +127,9 @@ Item {
             onClicked: {
                 isTramming = true;
                 buttonSelected = -1;
-                X1Plus.sendGcode(X1Plus.GcodeGenerator.macros_tramming(X1Plus.GcodeGenerator.TRAMMING_STEP.PREPARE));
+                const trammingGcode = gcodeLibrary.Tramming.prepare()
+                X1Plus.sendGcode(trammingGcode,seq_id);
+                console.log(trammingGcode);
```
jwise, line 132: do we need to log this?

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -55,106 +13,48 @@ function createGcode(command, params = {}) {
...
+/** 
+ * skew compensation 
+ * usage: gcode = GcodeGenerator.M1005.xy(84.99, 84.98)
+ * or
+ * gcode = GcodeGenerator.M1005.i(0.00015)
+*/
+var M1005 = { 
+    xy: (x,y) => `M1005 X${x} Y${y}\n`,
+    i: (i) => `M1005 I${i}\n`,
+    save: () => M500,
```
jwise, line 57: This is an error.  There is no variable named M500.

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: > This looks like a straightforward linear interpolator. Should it be in GcodeGenerator or in whatever timer is actually doing the LERP?

There's no timer used for the speed ramp feature. It uses an onLayerNumChanged signal in QML. Anyway I think we could find a better place for it than here, since it's not generating gcode

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -55,106 +13,48 @@ function createGcode(command, params = {}) {
...
+}
+
+
+/** 
+ * skew compensation 
+ * usage: gcode = GcodeGenerator.M1005.xy(84.99, 84.98)
+ * or
+ * gcode = GcodeGenerator.M1005.i(0.00015)
+*/
+var M1005 = {
```
jwise, line 54: One thing in the back of my head is that it is sort of weird and inconsistent that in some places we use human-readable names and in some places we just use G-code commands.  Particularly when we start doing things like `GcodeGenerator.M1002.gcode_claim_action` later, it makes me think that this should be, like, `GcodeGenerator.skewCompensation.xy` -- i.e., either everything should be Gcode, or everything should be human readable, but not something in between.

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -230,61 +145,73 @@ function M975(enabled = true) {
...
+
+/**
+ * Generates G-code based on the print speed level.
+ * 
+ * @param {number} speedPercentage The desired print speed as a percentage of the normal speed (100%).
+ *                                 Accepts: 50 (Silent), 100 (Normal), 125 (Sport), 166 (Luda)
+ * @returns {string} The G-code string to set print speed
+ */
+function speed(speedPercentage) {
+    if (speedPercentage <30 || speedPercentage > 180){
```
jwise, line 172:
```suggestion
    if (speedPercentage < 30 || speedPercentage > 180){
```

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -300,22 +227,19 @@ function M970({axis, a, f_low, f_high, h, k})
...
-}
-
-function M400(sec = 0) /* pause */
-{
-    return sec > 0 ? `M400 S${sec}\n` : `M400\n`;
+/* pause */
+var M400 = {
+    pause: (t) => `M400 S${t}\n`, //delay _ seconds
+    M400:()=> "M400\n", //wait for last gcode command to finish
+    pause_user_input:()=>"M400 U1\n", //pause until user presses "Resume"
```
jwise, line 238:
```suggestion
    M400: () => "M400\n", //wait for last gcode command to finish
    pause_user_input: () => "M400 U1\n", //pause until user presses "Resume"
```

bbl_screen-patch/patches/printerui/qml/printer/BedTramming.qml
```diff
@@ -14,7 +14,9 @@ Item {
     property bool emulating: X1Plus.emulating != ""
     property var buttonSelected: -1
     property bool isIdle: task.stage < PrintTask.WORKING && X1Plus.DDS.gcodeAction() != 254
-
+    property var gcodeLibrary: X1Plus.GcodeGenerator;
+    property var seq_id: 0
```
jwise, line 18: Why was this added?  This is a const of 0, isn't it?

bbl_screen-patch/patches/printerui/qml/settings/ConsolePage.qml
```diff
@@ -13,104 +13,202 @@ import ".."
...
-                ]
-    property var cmds: [" $ ","  ( )  "," ` ", "  { }  ","  |  ","  -  ","  &  ","  /  ", "reboot","awk ","cat ", "chmod ","chown ", "chroot", "cp ","date -s ", "dd ", "df ", "echo ","grep", "head ","ifconfig", "iptables ", "kill ","killall ","ln -s","ls -l ","mount ","mv ","pgrep ","pidof","ping -c 1","poweroff","print ","ps aux ", "ps -ef ", "pwd", "remount", "rm ", "sed","sort","tar","test","touch ", "uname -a"]
+    property int historyIndex: 0
+
+    property bool gcodeConsole: DeviceManager.getSetting("cfw_default_console",false);
+    property alias inputText: inputTextBox.text;
+    property bool printing: PrintManager.currentTask.stage >= PrintTask.WORKING
+    property bool ignoreDialog: false;
+    property var gcodeLibrary: X1Plus.GcodeGenerator
+    property var gcodeCommands: [
```
jwise, line 23: The name -> action mapping is a huge win here!  Thanks for this.

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jwise: Yeah, whatever the source is, I think the LERP belongs there

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: I see your point, and it's a valid concern. 

But this is not something I would be putting up a PR for if I had any question about. The printer only reports even numbered percentages, but I've tested every single % from 30 to 180% on about 40 hours of prints. Also Wolf_with_sword added the same trendlines and a speed slider to his HA workflow so this is already extremely well tested. 

I have a writeup of the interpolation and a comparison of Bambu's parameters with curve fit data, which I can add to the wiki in a format that makes sense. I don't know what the answer to this question should be though. We could fix the values for the 4 different speed levels, but there isn't a way to fit all of the points easily. I'm willing to modify it if you think that's best, just need some suggestions.

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: This is their typo. I will fix ours property name

bbl_screen-patch/patches/printerui/qml/settings/ConsolePage.qml
```diff
@@ -270,7 +355,19 @@ Item {
...
+        if (cmdHistory.length === 0) return;
+
+        if (up) {
+            historyIndex = (historyIndex + 1) % cmdHistory.length;
+        } else {
+            historyIndex = (historyIndex - 1 + cmdHistory.length) % cmdHistory.length;
+        }
+
+        inputText = cmdHistory[historyIndex];
+    }
```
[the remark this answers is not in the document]
jphannifan: This is an old feature! See https://github.com/X1Plus/X1Plus/blob/d5f2386fb766b95dfb90b24ac5a7991027d29135/bbl_screen-patch/patches/printerui/qml/settings/ConsolePage.qml#L377

You are right that the conditional is unnecessary. I will modify this

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: no it's a function. fixed

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: I had the same thought but I wasn't sure what was most intuitive. What made the most sense to me was naming each function by their command name because everyone knows what `M400` means but they're less likely to know what `saveParameters` means. I agree there should be a stricter naming convention here

bbl_screen-patch/patches/printerui/qml/settings/ConsolePage.qml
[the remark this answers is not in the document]
jphannifan: > Sigh. I really wish this new feature were not buried in an already large PR. I really really really want PRs to be one thing at a time so that they are more reviewable. We have had history-related bugs before, so this actually needs to be looked at... and it's really hard to do that when there are an enormous amount of other diffs to the same file.
> 
> In this specific case, if `up` is always `true`, the conditional should be removed -- code that is never exercised (and, therefore, never tested!) should not be added.

Why don't I pull this into a different PR?

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: I do agree something needs to change, I'm just not sure what yet. I'll think on it

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: We were publishing this command because it was in the sequence from /usr/etc/print/auto_cali_for_user_param.gcode. The gcode files from this folder still do not seem to have been shared publicly online yet, because some commands have no google search results. 

And for that same reason that means we really should add this back.

PR #203, opened 2024-04-24
Allow creating cron jobs

m1guelpf:
Allow users to define cron jobs by placing a crontab at `x1plus/printers/<serialnumber>/` on their SD card. Only starts `crond` if the file is found.

Wasn't sure which prefix to add to the `init.d` entry, so I went with `S99`, let me know if this should be something else. Also, this should probably be added to the wiki, but I couldn't find how to add that to the PR.

### DIFF
--- a/images/cfw/etc/init.d/S99crontab
+++ b/images/cfw/etc/init.d/S99crontab
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-DEVICE_SN=$(cat /proc/cmdline | xargs -n1 | grep "bbl_serial=" | sed "s|bbl_serial=||")
+DEVICE_SN=$(bbl_3dpsn 2> /dev/null)
 
 CRONTAB_FILE="/mnt/sdcard/x1plus/printers/$(DEVICE_SN)/crontab"
 
--- a/images/cfw/etc/init.d/S99crontab
+++ b/images/cfw/etc/init.d/S99crontab
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+DEVICE_SN=$(cat /proc/cmdline | xargs -n1 | grep "bbl_serial=" | sed "s|bbl_serial=||")
+
+CRONTABS_DIR="/mnt/sdcard/x1plus/printers/$(DEVICE_SN)/crontab"
+
+start() {
+    if [ -f "$CRONTABS_DIR" ]; then
+        mkdir -p /var/spool/cron/crontabs
+        ln -sf "$CRONTABS_DIR" /var/spool/cron/crontabs/root
+        echo "Starting crond..."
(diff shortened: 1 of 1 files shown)


images/cfw/etc/init.d/S99crontab
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+
+DEVICE_SN=$(cat /proc/cmdline | xargs -n1 | grep "bbl_serial=" | sed "s|bbl_serial=||")
```
jwise, line 3: This will work, though this shenanigan is needed only at initial boot before we have most of the security service started.  I think the more idiomatic way to do this after filesystems are mounted is `$(bbl_3dpsn)`.

images/cfw/etc/init.d/S99crontab
```diff
@@ -0,0 +1,46 @@
...
+
+CRONTABS_DIR="/mnt/sdcard/x1plus/printers/$(DEVICE_SN)/crontab"
+
+start() {
+    if [ -f "$CRONTABS_DIR" ]; then
+        mkdir -p /var/spool/cron/crontabs
+        ln -sf "$CRONTABS_DIR" /var/spool/cron/crontabs/root
```
jwise, line 10:
```suggestion
CRONTAB_FILE="/mnt/sdcard/x1plus/printers/$(DEVICE_SN)/crontab"

start() {
    if [ -f "$CRONTAB_FILE" ]; then
        mkdir -p /var/spool/cron/crontabs
        ln -sf "$CRONTAB_FILE" /var/spool/cron/crontabs/root
```
perhaps?

images/cfw/etc/init.d/S99crontab
m1guelpf: I see other places also check `/oem/device/sn` before calling `bbl_3dpsn`. Should I include that here as well?

jwise:
S99 is probably fine.  Thanks for the PR!  Eyeballing it, I'm OK with this, though I have not tested it on my own printer.

images/cfw/etc/init.d/S99crontab
jwise: Those are only necessary for the install/kexec environment, which needs to be compatible with old base firmwares that may not have a working `bbl_3dpsn`.  I think `bbl_3dpsn` is good enough here.

jwise (approved):
This is great!  Thanks for the PR!

m1guelpf:
not 100% sure what happened with the build, maybe it's related to GitHub not using secrets when building from forks?

jwise:
yeah, that's correct.  how annoying.  I will rambo it past CI

PR #205, opened 2024-04-24
symlink python

bani6809:


jphannifan (approved):

jwise:
this is fine-ish, though here is one thing I was thinking earlier today: it would really be nice, I think, to just add /opt/python3/bin to the PATH (and, also, put x1plus tools in, like, /opt/x1plus, rather than /opt directly -- and then have /opt/x1plus/bin on the PATH).  WDYT?

PR #207, opened 2024-04-24
ConsolePage command history feature added

jphannifan:
Prior to 1.0 we had a "command history" feature on ConsolePage.qml which allowed you to pull up previously published commands. The reason it was removed was because there was an issue with the tap gesture that was used to access the feature. It's now a button called "History" in the bottom pane. Switching between Gcode and shell or vice versa clears the history.

Maybe this would look better as an icon instead of "History"? It's not exactly obvious what a button called "History" does.

I've also adjusted the width property of the TextArea to fix a binding loop error. I've tested it and the width remains the same and the error no longer appears.

### DIFF
--- a/bbl_screen-patch/patches/printerui/qml/settings/ConsolePage.qml
+++ b/bbl_screen-patch/patches/printerui/qml/settings/ConsolePage.qml
@@ -234,7 +234,7 @@ Item {
             clip: true
             TextArea {
                 id: outputTextArea
-                width: 1130 - 56
+                width: 1130 /* parent.width, but avoiding a binding loop */ - 56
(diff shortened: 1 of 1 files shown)


riptidewave93 (approved):

bbl_screen-patch/patches/printerui/qml/settings/ConsolePage.qml
```diff
@@ -234,7 +234,7 @@ Item {
             clip: true
             TextArea {
                 id: outputTextArea
-                width: parent.width - 56
+                width: 1130 - 56
```
bani6809, line 237: y hardcode?

bani6809 (approved):
seems to work

PR #210, opened 2024-04-25
add docker scripts

bani6809:


jphannifan (approved):

riptidewave93:
Instead of scripts, maybe these should be a makefile command? make build-docker for example?

bani6809:
> Instead of scripts, maybe these should be a makefile command? make build-docker for example?

make assumes you even have make/gcc/etc installed, which a barebones linux install may not have, WSL does not for sure. the docker is guaranteed to.

docker-run.sh
```diff
@@ -0,0 +1 @@
+sudo docker run --name x1plus -it -v `pwd`:/work x1plus
```
jwise, line 1: should we have users run the prebuilt docker image instead of building it themselves?  then we know they'll get the same thing we have in CI

docker-clean.sh
```diff
@@ -0,0 +1 @@
+sudo docker system prune -a
```
jwise, line 1: hm I am not sure I want this to be an x1plus script.  this has far wider reaching consequences than just x1plus!

jwise:
I think I agree that this is not a `make` thing but also I'm not sure we need to provide these scripts at all.  if you don't have make installed you probably don't have docker installed either :)

bani6809:
> I think I agree that this is not a `make` thing but also I'm not sure we need to provide these scripts at all. if you don't have make installed you probably don't have docker installed either :)

i dont have make installed on the host because i dont want a minimalist host. build env is inside the container not outside.

jphannifan (approved):

riptidewave93:
Personally I feel these are unneeded with good documentation, but that's just me. :man_shrugging:

riptidewave93 (approved):

docker-run.sh
fonix232: I'd say yes, a prebuilt image is preferred - not everyone might have a powerful machine on hand to build the base image fast.

I'd probably also commit the /work directory into the image so that no git checkout is necessary for basic work/build. That way it's an atomic, "guaranteed to work" image.

Another note I'd add here is running the container in detached mode. With the current commands (especially the ones provided in the README), for each command, a new container is spun up then shut down once the command completes. With a detached container, just by referring to the name and using `exec` instead of run, a lot of resources can be saved.

For example the current README approach is:

```
$ docker run -u `id -u` -v `pwd`:/work x1plusbuild bash -c 'git config --global --add safe.directory /work'
$ docker run -u `id -u` -v `pwd`:/work x1plusbuild make scripts
$ scp scripts/getkey root@bambu:/tmp
$ ssh root@bambu /tmp/getkey >> localconfig.mk
$ docker run -u `id -u` -v `pwd`:/work x1plusbuild make
```

Which could be modified to:

```
$ docker run -d --name x1plus -u `id -u` -v `pwd`:/work x1plusbuild 
$ docker exec -it x1plus bash -c 'git config --global --add safe.directory /work'
$ docker exec -it x1plus make scripts
$ scp scripts/getkey root@bambu:/tmp
$ ssh root@bambu /tmp/getkey >> localconfig.mk
$ docker exec -it x1plus make
$ docker stop x1plus
// Later on, the same container can be spun up with
$ docker start x1plus
```

PR #211, opened 2024-04-25
ci: logging and file improvements

riptidewave93:
* Silence our wget of the BBL firmware, it's over 50% of our build logs
* gitignore our ts file, since it's pulled on build and not committed to the repo.


jwise (approved):
Agreed.  Thanks.  The whole build system sucks rocks and leaves files everywhere but if we're going to do that we might as well at least gitignore them...

PR #212, opened 2024-04-25
fix: add buildtimestamp to ota file

riptidewave93:
TL;DR it's hard to compare builds with our current naming scheme, so let's add a build timestamp that can be used for this compare.


jwise (approved):
LGTM.  Thanks!

PR #216, opened 2024-04-26
fix: move jq to prebuilt and add to internal-fs

riptidewave93:
This is pre-req work for once x1plusd is ready. But TL;DR, since we read from our settings.json direct in the future, we will need JQ in multiple spots. So now we just reference it from prebuilt, where it should have been this entire time.


jwise (approved):
LGTM, and this also gives me an idea for how to fix the mksquashfs hell I am in

PR #217, opened 2024-04-27
Adding Web HMI. Lightweight Flask webserver that runs on X1Plus itself.

christophergault198:
Working flask web HMI for X1Plus. Served on port: `5001`. Currently returns current toolhead and FLC images with multiple printer commands and currently working on a ML model with scooterthedev. Yes, `dark mode' is included. @jphannifan @riptidewave93
image


christophergault198:
Build did not append any of the folders as expected including `/images/cfw/etc/init.d/S98webhmi` ,`/images/cfw/usr/etc/system/webhmi` folder. Also missing venv folder from build, need to include.

scooterthedev:
I believe it should be pushed when we have a finished product, so we are not putting in a pr everytime we add a small feature. It hasn't been fully tested yet on device, and I'm worried that it's not going to work as expected when it's pushed to 1.2. Lets take some more time to finalize everything, and make sure that there are no bugs, or anything that would conflict other stuff from working ie. running out of RAM, using a port that's is meant to be used for something else etc. @christophergault198

christophergault198:
> I believe it should be pushed when we have a finished product, so we are not putting in a pr everytime we add a small feature. It hasn't been fully tested yet on device, and I'm worried that it's not going to work as expected when it's pushed to 1.2. Lets take some more time to finalize everything, and make sure that there are no bugs, or anything that would conflict other stuff from working ie. running out of RAM, using a port that's is meant to be used for something else etc. @christophergault198

Noted, mainly intended to bring awareness outside. This is why build has not been reported. Will use more appropriate channels.

christophergault198:
Looking for broader ideas here. Otherwise should Close?

christophergault198:
Closing due to needed improvements and review.

PR #218, opened 2024-04-27
SD logging for syslog_shim logger

jphannifan:
syslog_shim.py checks to see if SD logging is enabled, but it still saves its own log to /tmp/ no matter what.

I've fixed that here. 

Also added regex to syslog_shim.py so it will capture skew factor data:

```
[MCU][BMC]M1005:current XY_comp_ang = 0.00110
[MCU][BMC]M1005:new XY_comp_ang = -0.00097
```

Note: These changes and syslog_shim.py as a whole will become mostly irrelevant as soon as we set up syslog-ng

### DIFF
--- a/images/cfw/opt/syslog_shim.py
+++ b/images/cfw/opt/syslog_shim.py
@@ -7,7 +7,17 @@ from collections import namedtuple
 from logger.tail import TailLog
 import dds
 
-syslog_log = CustomLogger("Syslog parser", "/tmp/x1plus_data.log", 500000, 1)
+# Log setup - Check if SD logging is enabled
+# Start up our own logger (to sd card if SD logging, /tmp/ if not)
+log_path = (
+    "/mnt/sdcard/log/"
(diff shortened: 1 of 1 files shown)


images/cfw/opt/syslog_shim.py
```diff
@@ -7,7 +7,17 @@
 from logger.tail import TailLog
 import dds
 
-syslog_log = CustomLogger("Syslog parser", "/tmp/x1plus_data.log", 500000, 1)
+# Log setup - Check if SD logging is enabled
+# Start up our own logger (to sd card if SD logging, /tmp/ if not)
+log_path = (
+    "/mnt/sdcard/log/"
```
jwise, line 13: I was going to complain that this should be in `/mnt/sdcard/x1plus/printers/NNNNN/logs/` but it is not a regression over what we have, looking further below

jwise (approved):
Yeah, I agree, and I also agree that this should be revisited in general in the near future.

PR #219, opened 2024-04-27
use tar2sqfs instead of gensquashfs

jwise:
gensquashfs had a bunch of misbehavior in the past, including generating duplicates of /opt/python/bin.  Instead, we do something truly terrible, involving generating a tar file with what we want, and then packing that.  This sucks but the only thing that sucks worse is what we had, I guess.

At some point the tarfile generation stuff can be refactored out, once we have multiple images.  But this is fine for now and unblocks what I was *actually* trying to do with my night.


images/cfw/sbin/iwconfig
bani6809, line 1: wot dis?

bani6809 (approved):
seems to work

PR #221, opened 2024-04-27
add boot debugging feature

bani6809:
kexecui boot menu option to redirect init output to /dev/tty0 so we have a chance to debug this:

https://discord.com/channels/1041781310299062282/1184979423540285540/1231594286504607847


riptidewave93 (approved):
LGTM but I recommend a review from @jwise since this is touching the initramfs

jphannifan (commented):
This looks good to me. We may eventually need to make kexec_uis listview scrollable if we decide to add any more options.

jphannifan (approved):

jwise (approved):
I have no objection to the `init` and `boot` parts (though it would be nice to have consistent indentation in `init`).  I think also traditionally there is some kind of `shift` thing that you do to get the `/proc/cmdline` in `init` but this is fine.

That said, I kind of do object to the "Start X1Plus in debug mode" menu item.  I guess it's fine because it doesn't do anything bad, but generally exposing a button for users to tap on that makes their experience worse is a bad idea, especially because literally every single 3D printer user thinks they are an expert and therefore they enable every expert mode they can find without regard for the consequences.  So if "debug mode" is just logging at init, that's fine, but if it starts doing anything more painful to users later (i.e., slowing things down with logging?), it will be necessary to make this much more intentional.

PR #224, opened 2024-04-28
Use the correct MD5 for the 00.00.32.17 squashfs.

jwise:
Fixes an issue in which the installer would always regenerate the squashfs, even if it had a good one there.

Oops.  My b.


bani6809 (approved):
tested, confirmed matches.

riptidewave93 (approved):

PR #225, opened 2024-04-28
makefile cleans python, add python symlinks, add profile.d/x1plus.sh

bani6809:


images/cfw.py
```diff
@@ -75,6 +75,9 @@ def globfiles(fromdir, todir, eatlinks = []):
 symlink('/etc/localtime', '/usr/share/zoneinfo/Etc/UTC')
 symlink('/var/run/utmp', '/dev/null')
 symlink('/usr/sbin/mount.smb3', '/usr/sbin/mount.cifs')
+symlink('/usr/bin/python', '/opt/python/bin/python3.10')
+symlink('/usr/bin/python3', '/opt/python/bin/python3.10')
+symlink('/usr/bin/python3-config', '/opt/python/bin/python3.10-config')
```
jwise, line 80: are these necessary if we have /opt/python/bin on the PATH?  anyone who wants to have a python script should `#!/usr/bin/env python` anyway

bani6809: unless there is harm in doing so, why omit it?

also...

```
unix@kiyone:~/X1Plus$ find . -type f -name "*py" -print0 | xargs -0 grep "bin/python" | grep -v site-packages
./images/cfw/opt/x1plusd.py:#!/opt/python/bin/python3
./images/cfw/opt/python/lib/python3.10/cgi.py:#! /usr/local/bin/python
./images/cfw/opt/python/lib/python3.10/cgi.py:# NOTE: the above "/usr/local/bin/python" is NOT a mistake.  It is
./images/cfw/opt/x1plus-test-settings.py:#!/opt/python/bin/python3
```

jwise: I guess there is no harm in doing so.  though we use `/opt/python/bin/python3` because we do not have it on the path to begin with!

jwise (approved):
my name is joshua and I approve of this message

PR #227, opened 2024-04-30
add debug_wifi.sh script

bani6809:


riptidewave93 (approved):

jphannifan (commented):
Works for me. I think we could use error handling in a few more places that would help us debug the installer, but this is a good start.

jphannifan (approved):

jwise (approved):
This is not how I would implement it -- I probably would write this as follows:
```python
DEBUGS = [ ('dmesg', 'dmesg'), ('lsmod', 'lsmod'), ('/config/wifi/config.txt', 'cat /config/wifi/config.txt'), ...]
for descr, cmd in DEBUGS:
  print("==============")
  print(descr)
  print("==============")
  os.system(cmd)
  print("")
```

but this also works.

PR #228, opened 2024-04-30
start using the new standard for settings key, and migrate image locations to x1plus settings

jwise:
We decided it is easier to name things with dotted settings.  So we change `locktype` to `lockscreen.mode` and `passcode` to `lockscreen.passcode` (and fix the migration logic associated...).

We also migrate `cfw_lockscreen_image` to `lockscreen.image`, `cfw_home_image` to `homescreen.image`, and `cfw_print_image` to `homescreen.image.printing`.

As a bonus of this now being Dbus connected and binding connected, you can now change the image without restarting bbl_screen.  Try, for instance:

`./x1plus-test-settings.py homescreen.image.printing=/opt/scooter_profile.jpg`

(after putting a picture of my former cat in /opt), or:

`./x1plus-test-settings.py homescreen.image.printing=None`

to return to the default.  The screen will change without a reboot!

### DIFF
--- a/bbl_screen-patch/patches/printerui/qml/main/Home2Page.qml.patch
+++ b/bbl_screen-patch/patches/printerui/qml/main/Home2Page.qml.patch
@@ -20,7 +20,7 @@
      property var defaultImage: DeviceManager.getSystemSetting("device/images_path")
      property bool printPrepare: task.state < PrintTask.RUNNING && task.state > PrintTask.IDLE
 +    property var serialNo: DeviceManager.build.seriaNO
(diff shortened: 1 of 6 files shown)


bani6809:
there should be a maintained document somewhere listing all the keys and their purpose.

jwise:
I agree.

jwise:
I wrote a wiki page: https://github.com/X1Plus/X1Plus/wiki/Configuration-settings

bani6809 (approved):
seems to work

jphannifan (commented):
looking good.

It would be cool if we took this a step further and had settings "profiles" and you can change users on the lockscreen. I don't think we're quite ready for that yet but we might revisit it in the future

PR #230, opened 2024-05-01
Generate a correct brcmfmac43430-sdio.txt on boot, and include the regulatory.db as well.

jwise:
Ludacris may mention that he is "worldwide, bitch, act like y'all didn't know", but clearly not: he mentions only area codes like 770, 404, 418, 202, etc.  We, on the other hand, *are*, worldwide, though we previously acted like we didn't know; we must deal also with `US` and DE, and especially must avoid `ALL`.  The `bcmdhd` driver grabbed a `ccode` as an override from config.txt, but `brcmfmac` takes the nvram.txt on blind faith.  We update the nvram.txt data with the config.txt data, and for good measure to keep the kernel happy, we include a `regulatory.db` as well.

This should fix issues with channels 12 and 13 on European printers.


bani6809 (approved):
it is not less worse than status quo, and fixes the channel >11 issue for at least one DE user.

PR #235, opened 2024-05-04
Toggle for SD logging

jphannifan:
Enough users have been using this feature by now that I think we need a toggle for it. 

I think it's best if we keep this an undocumented feature, but it will make our jobs easier because we don't have to explain how and where to touch a file.


jwise (commented):
Isn't this replaced by x1plusd and x1plusd-settings?

jphannifan:
> Isn't this replaced by x1plusd and x1plusd-settings?

oof you're right

PR #236, opened 2024-05-04
feat: add ota engine and improved x1plusd

riptidewave93:
* Update settings with defaults, add boot flag file migrations
* Add our intial OTA check Class & dbus interface
* Move away from flag files in all init scripts


jwise:
The migration stuff seems good.  I think the OTA stuff could use another pass?  I'll comment inline here but I sort of think these should be split into two PRs so we can merge the migration stuff while we go back and forth on the OTA stuff.

images/cfw/etc/init.d/S01logging
```diff
@@ -2,6 +2,8 @@
 #
 # Start logging
 
+. /usr/share/x1plus/functions.sh
```
jwise, line 5: I think this is a good improvement.  One thing I am thinking is that we probably should be more disciplined about where on the filesystem we are dropping stuff.  I had originally been thinking of migrating a bunch of our tools to `/opt/x1plus/lib` and `/opt/x1plus/bin` and such.  I guess I could be persuaded that `/usr/share/x1plus` is a reasonable place too (or `/usr/x1plus/libexec/functions.sh` or something?).

I'm not sure I really have it in me to block this PR on that, but I think the day of reckoning for our paths is coming sooner rather than later.  So if you feel like it it might be nice for this PR to start whatever the New System of Paths is going to be.  My vote for this one is `/opt/x1plus/libexec/functions.sh`.

images/cfw/etc/init.d/S01logging
```diff
@@ -10,7 +12,7 @@ start() {
 	printf "Starting logging: "
 
 	# If persist is enabled, log to card
-	if [ -f "/mnt/sdcard/x1plus/printers/$DEVICE_SN/logsd" ]; then
+	if x1p_get_setting boot.sdcard_syslog logsd; then
```
jwise, line 15: Yes this is so darn much better.

images/cfw/etc/init.d/S02backup_emmc
```diff
@@ -2,9 +2,25 @@
...
+clear_flag() {
+    # This is here instead of functions.sh because ONLY backup_emmc should 
+    # clear it's own flag, as it runs before x1plusd, so it needs to handle
+    # the flag reset on it's own if we are moved to settings.json
+    [ -f "/mnt/sdcard/x1plus/printers/${DEVICE_SN}/dump-emmc" ] && rm /mnt/sdcard/x1plus/printers/${DEVICE_SN}/dump-emmc
+    if [ -f "/mnt/sdcard/x1plus/printers/${DEVICE_SN}/settings.json" ]; then
+        TMPSETTINGS=$(mktemp)
+        jq ".\"boot.dump_emmc\" = false" /mnt/sdcard/x1plus/printers/${DEVICE_SN}/settings.json > "${TMPSETTINGS}"
+        mv "${TMPSETTINGS}" /mnt/sdcard/x1plus/printers/${DEVICE_SN}/settings.json
+    fi
```
jwise, line 23: This is good documentation for this.  I agree with this.

images/cfw/etc/init.d/S02backup_emmc
```diff
@@ -14,30 +30,27 @@ dump_emmc() {
...
 
-    # Get SN from cmdline
-    DEVICE_SN=$(cat /proc/cmdline | xargs -n1 | grep "bbl_serial=" | sed "s|bbl_serial=||")
-
     # Now that sd is mounted, do we need to do an emmc dump?
-    if [ -f /mnt/sdcard/x1plus/printers/$DEVICE_SN/dump-emmc ]; then
+    if x1p_get_setting boot.dump_emmc dump-emmc; then
         msg "Attempting to dump emmc as requested..." # Progress spinner sits ontop of this but that's fine
-        if [ -f /mnt/sdcard/x1plus/printers/$DEVICE_SN/emmc.bin ]; then
+        clear_flag # Clear now, so we unset even if there's no free disk space to do the dump
```
jwise, line 36: ...though maybe I would call it `_clear_flag`, since looking at this I go like "huh what is a bash command named clear_flag?".

images/cfw/etc/init.d/S70x1plus_sshd
```diff
@@ -6,6 +6,7 @@ HOST_KEYFILE="${CONFIG_DIR}/dropbear_ecdsa_host_key"
 PIDFILE="/var/run/x1plus_sshd.pid"
 
 update_rootpw() {
+  # TODO: Migrate to x1p_get_setting once bbl_screen moves over to settings.json
```
jwise, line 9: Agree.  Or x1plus_sshd might be controlled by x1plusd, so it will react to settings changes.

images/cfw/opt/x1plusd.py
```diff
@@ -1,116 +1,282 @@
...
+
+                if not rv:
+                    logger.warning(f"{method} -> NoMethod")
+                    await self.router.send(
+                        new_error(msg, "x1plus.x1plusd.Error.NoMethod")
+                    )
+                    continue
+
+                logger.debug(f"{method}({arg}) -> {rv}")
+                await self.router.send(new_method_return(msg, "s", (rv,)))
```
jwise, line 105: It would be nice to refactor the 'Dbus RPC catcher', since this otherwise repeats the thing below, and it would be nice not to have that coded xeroxed.

images/cfw/opt/x1plusd.py
```diff
@@ -1,116 +1,282 @@
...
+            ) as fh:
+                x1p_settings = json.load(fh)
+        except FileNotFoundError:
+            logger.warning("OTA can't find settings.json, assuming OTAs are disabled!")
+            x1p_settings = {"ota.enable": False}
+
+        # Do we have OTAs enabled? If not, just return current status
+        if not x1p_settings.get("ota.enable", False):
+            logger.info("OTA check is disabled, skipping check!")
+            return json.dumps({"ota": ota_response})
```
jwise, line 133: We should 'know' this -- just reach into the settings daemon's state to ask how it's feeling.

images/cfw/opt/x1plusd.py
```diff
@@ -1,116 +1,282 @@
...
+        try:
+            # Update check timestamp first
+            self.last_check_timestamp = datetime.datetime.now().timestamp()
+            r = requests.get(self.ota_url, timeout=5)
+            self.last_check_response = r.json()
+        except Exception as e:
+            logger.error(f"Exception calling OTA URL! Error of: {e}")
+            # we Timed out, or hit other error with requests
+            self.last_check_error = True
+            return json.dumps({"ota": ota_response})
```
jwise, line 145: We shouldn't do anything synchronous on the Dbus thread -- that will block every other Dbus user in x1plusd.  It's ok to use aiohttp here, but not asyncio.  It's also okay to have a background thread. 

There's also another problem, though, which is that a Dbus API call should be 'fast', because it blocks the UI thread for bbl_screen.  Maybe the API is, like:

```
x1plus.ota.GetStatus({}) -> {status structure}
x1plus.ota.CheckNow({}) -> {} # starts the ota background task checking, returns immediately
```

And then as things change in the X1Plus OTA world, it emits signals (the way the settings service does):
```
x1plus.ota.Status({status structure})
```

And the status structure is something like:
```javascript
{
  "check_enabled": ...,
  "check_interval": ...
  "check_timestamp": timestamp,
  "check_status": "OK" /* or "CHECKING", or "CHECK_FAILED" */,
  "version_ota": { version info... },
  "version_current": { version info... },
  "download_status": { some structure? },
}
```

or something?

images/cfw/opt/x1plusd.py
```diff
@@ -1,116 +1,282 @@
...
+            self.settings = self._migrate_old_settings()
             self._save()
-    
+
+    def _migrate_old_settings(self):
+        """
+        Used to migrate init.d flag files to our new json on first run.
+        All bbl_screen settings MUST BE MIGRATED BY bbl_screen!
+        """
+        defaults = copy.deepcopy(SettingsService.DEFAULT_X1PLUS_SETTINGS)
```
jwise, line 191: maybe it should just be `{}` as the default no matter what?  the default is False after all for absent things

images/cfw/opt/x1plusd.py
```diff
@@ -1,116 +1,282 @@
...
+            defaults.update({"boot.dump_emmc": True})
+            Path(f"{self.settings_dir}/dump-emmc").unlink(missing_ok=True)
+        # syslog to sd
+        if os.path.exists(f"{self.settings_dir}/logsd"):
+            defaults.update({"boot.sdcard_syslog": True})
+            Path(f"{self.settings_dir}/logsd").unlink(missing_ok=True)
+        # performance logging (debugging)
+        if os.path.exists(f"{self.settings_dir}/perf_log"):
+            defaults.update({"boot.perf_log": True})
+            Path(f"{self.settings_dir}/perf_log").unlink(missing_ok=True)
```
jwise, line 208: DRY... how about something like...
```suggestion
        for fname,flag in [("quick-boot","boot.quick_boot"), ("dump-emmc","boot.dump_emmc"), ("logsd", "boot.sdcard_syslog"), ("perf_log", "boot.perf_log")}:
          if os.path.exists(f"{self.settings_dir}/{fname}")):
              defaults[flag] = True
              Path(f"{self.settings_dir}/{fname}").unlink(missing_ok = True)
```

internal-fs/opt/kexec/check_kexec
```diff
@@ -16,18 +16,28 @@ else
     BBL_SN=$(bbl_3dpsn 2> /dev/null)
 fi
 
+# Set quickboot var
+QUICKBOOT_FLAG="false"
+if [ -f /mnt/sdcard/x1plus/printers/${BBL_SN}/settings.json ]; then
+    # Rely on settings.json if we can, default to false
+    QUICKBOOT_FLAG=$(/opt/kexec/jq ".\"boot.quick_boot\" // false" /mnt/sdcard/x1plus/printers/${BBL_SN}/settings.json)
```
jwise, line 23: ... and rm the flag

internal-fs/opt/kexec/check_kexec
```diff
@@ -16,18 +16,28 @@ else
...
     for x1p in $(ls -1 /mnt/sdcard/*.x1p | sed -e 's|/mnt/sdcard/||g' | sed -e 's|.x1p||g'); do
         # Doesn't exist, bomb out
         if [ ! -d "/mnt/sdcard/x1plus/kernel/${x1p}" ]; then
-            QUICK_BOOT=1 # Disable
+            NO_NEW_X1P=1 # Disable
         fi
     done
     # If we are good to go, quick boot.
-    if [ "${QUICK_BOOT}" -eq "0" ]; then
+    if [ "${NO_NEW_X1P}" -eq "0" ]; then
```
jwise, line 40: the inverted logic sense of this 1 vs. 0 makes me confused but I guess we did not really change it here otherwise.

images/cfw/opt/x1plusd.py
riptidewave93: Maybe we define a parent class that defines this and we inherit it for settings and OTA? We can think on it as I'm going to strip OTA from this PR and move it to its own draft PR after the settings logic is merged.

jwise: That’s kind of my thought, yeah.

riptidewave93:
For now im moving this PR to draft, and am making a new PR for all of the settings changes with improvements from the above comments.

images/cfw/opt/x1plusd.py
riptidewave93: idea is to also have a source of truth for defaults in one spot at file generation, vs defaults all over the place in bbl_screen.

internal-fs/opt/kexec/check_kexec
riptidewave93: Was doing 0/1 cuz bash return codes, but yes it's kinda confusing, ill leave for now

images/cfw/opt/x1plusd.py
riptidewave93: I mean, unless we wanna dbus across within the same app, feels dumb, esp since in our initial draft doc we said we should trust the settings file as the source of truth.

However, if you prefer this is dbus so we ensure ALL things go through the settings dbus interface, im not against that. :)

images/cfw/opt/x1plusd.py
riptidewave93: hmm, im not againt this, because you are right about async for the request. Any chance we can get aiohttp or grequests added to the python 3.12 PR and that becomes a merge dependency before OTAs?

images/cfw/opt/x1plusd.py
jwise: Well, what I mean by this is that we are in-process.  So you could, say, just ask the `SettingsService.settings` or something, no need for any DBus-to-DBus.  Right now that's hidden in `settings` but presumably we could have a `x1plusd` object that holds all of this.

images/cfw/opt/x1plusd.py
jwise: I was going to say that I think this can be orthogonal to 3.12, but it turns out that there are native components for aiohttp.  So it would be easier to do this simultaneously with 3.12, I guess.  Anyway, now that is on the jwise/python3.12 branch, feel free to check it out.

I wonder if we should build python cache files at install time / packaging time.  First startup is insanely slow.

riptidewave93:
Closed as https://github.com/X1Plus/X1Plus/pull/265 is improved

Issue #226, opened 2024-04-30
Unable to install X1Plus


micahfocht:
**Describe the bug**
Installer always fails at the "Running on-printer installation" step, particularly after asking to download 

**To Reproduce**
1) Walk through the installer application
2) Fail to download firmware on printer.

**Expected behavior**
I expect the installer to download the relevant firmware files.

**Screenshots/Photos**
![Screenshot 2024-04-29 at 9 01 22 PM](https://github.com/X1Plus/X1Plus/assets/15097414/32e7b047-6a45-4a21-bd9d-29968107e9da)

`Traceback (most recent call last):
  File "/userdata/x1plus/install.py", line 295, in <module>
    report_failure(f"Download failed.", e)
  File "/userdata/x1plus/install.py", line 75, in report_failure
    raise e
  File "/userdata/x1plus/install.py", line 280, in <module>
    with urllib.request.urlopen(basefw_update_url, capath="/etc/ssl/certs") as resp, \
  File "/userdata/x1plus/lib/python3.10/urllib/request.py", line 216, in urlopen
    return opener.open(url, data, timeout)
  File "/userdata/x1plus/lib/python3.10/urllib/request.py", line 525, in open
    response = meth(req, response)
  File "/userdata/x1plus/lib/python3.10/urllib/request.py", line 634, in http_response
    response = self.parent.error(
  File "/userdata/x1plus/lib/python3.10/urllib/request.py", line 563, in error
    return self._call_chain(*args)
  File "/userdata/x1plus/lib/python3.10/urllib/request.py", line 496, in _call_chain
    result = func(*args)
  File "/userdata/x1plus/lib/python3.10/urllib/request.py", line 643, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 403: Forbidden`

**External Links**
I have not seen discussion of this anywhere.  I saw similar discussion of slicers causing issues, but I ensured none were running and have what appears to be a different error.

**Additional Details:**
 - Printer type (X1/X1C/X1E): X1C
 - X1Plus version: I have attemped 1.0, 1.1 RC, and 1.1
 - AP firmware version: I'm not sure how to see all this on stock firmware.  It's entirely possible I'm just unaware.
 - AP hardware revision:
 - MC firmware version:
 - MC hardware revision:
 - Toolhead firmware version:
 - Toolhead hardware revision:
 - Computer vision model:

[...]

Issue #184, opened 2024-04-15
X1Plus custom firmware installation download failed


gogsi1:
**Describe the bug**
I installed X1 Plus 1.0 on 01.07.02.00, after that I installed update 1.0-7+gca7e1d4.x1 and wanted to install X1Plus1.1rc1, but it gives an error because in this mode the printer does not connect to wifi.

**To Reproduce**
1. I downloaded the firmware to the SD card
2. Logged into the x1plus installer
3. I chose the firmware to install X1Plus1.1rc1

**Screenshots/Photos**
![photo_2024-04-15_13-43-59](https://github.com/X1Plus/X1Plus/assets/36478778/8c42cb09-0972-4e2a-bf1f-cc5a1b94c02e)

**External Links**
It seems that I found one topic, but there is not much how to fix it there

**Additional Details:**
 - Printer type (X1):

riptidewave93:
Hello @gogsi1, can you confirm if your printer is in LAN or Cloud mode? If you can, please boot it into the official firmware (updater disabled) and confirm it is set to Cloud mode. Once done, please reboot and re-try installing the 1.1rc1 update.

If you continue to have issues, it's highly recommended to use the Discord for assistance. https://github.com/X1Plus/X1Plus/wiki#sounds-good-now-what

gogsi1:
> Hello @gogsi1, can you confirm if your printer is in LAN or Cloud mode? If you can, please boot it into the official firmware (updater disabled) and confirm it is set to Cloud mode. Once done, please reboot and re-try installing the 1.1rc1 update.
> 
> If you continue to have issues, it's highly recommended to use the Discord for assistance. https://github.com/X1Plus/X1Plus/wiki#sounds-good-now-what

Cloud mode is set

gogsi1:
> I had this same issue. After seeing this post I booted to official firmware (updater disabled) to check cloud access which was on. But I noticed in this mode the WiFi was not connected. So I connnected to WiFi while booted to official firmware then restarted. Now it downloads the base firmware just fine.

Thanks, I was able to update.

Turtletrumpet:
I run X1plus V1.0, I had this same issue. After seeing this post I booted to official firmware (updater disabled) to check cloud access which was >off<. Wifi was connected. I activate cloud access and power cycle the printer. I start the update process again and no download of base firmware is possible. I see on the wifi AP, that the printer is not connected. Please help. Thank you.

jphannifan:
[...]

PR #239, opened 2024-05-05
Bug fix: Kexec_ui installer download (issues 184, 226)

jphannifan:
This one was a two parter.

1) `HTTP Error 403: Forbidden` when kexec_ui tries to download an update (user agent issue)

2) kexec_ui race condition. wifi not enabled when needed

Added error handling for these and similar issues.

See Github issues [184](https://github.com/X1Plus/X1Plus/issues/184) and [226](https://github.com/X1Plus/X1Plus/issues/226) for more information on these errors.


installer/install.py
```diff
@@ -354,31 +354,19 @@ def download_firmware(update_url, dest_path):
...
+            time.sleep( pow( 2, retry ) if retry < 4 else 16 )
             retry += 1
         except Exception as e:
             report_interim_progress(f"An error occurred: {e} ... retry {retry}")
-            if retry < 4:
-                time.sleep( pow( 2,retry ) )
-            else:
-                time.sleep( 16 )
+            time.sleep( pow( 2, retry ) if retry < 4 else 16 )
             retry += 1
```
jphannifan, line 300: There is nothing stopping this from retrying in an endless loop if it can't download. Seems like we could break after a fixed number of retries

installer/install.py
```diff
@@ -268,42 +266,74 @@ def debug_wifi():
...
+        'User-Agent': f"Python-urllib/3.{sys.version_info.minor}"
+    }
+    
+    os.makedirs(os.path.dirname(dest_path), exist_ok=True)
+
+    report_interim_progress("Connecting...")
+    success = False
+    retry = 1
+    while success == False:
+        try:
```
jphannifan, line 282: This try block can be shortened

```
try:
     resp = requests.get(update_url, headers=headers, stream=True, verify="/etc/ssl/certs/")
     resp.raise_for_status()
except (requests.Timeout, requests.exceptions.HTTPError, requests.exceptions.RequestException) as e:
     report_interim_progress(f"{e.__class__.__name__}: {e} ... retry {retry}")
     time.sleep( pow( 2, retry ) if retry < 4 else 16 )
     retry += 1
except Exception as e:
     report_interim_progress(f"An error occurred: {e} ... retry {retry}")
     time.sleep( pow( 2, retry ) if retry < 4 else 16 )
     retry += 1
else:
     success = True
```

bbl_screen-patch/kexec_ui/printerui/image/wifi_0.svg
jwise, line 1: these have been removed from the root.qrc and should be removed from the PR too

bbl_screen-patch/kexec_ui/printerui/qml/BootOptionsPage.qml
```diff
@@ -15,7 +15,7 @@ Rectangle {
     width: 1280
     height: 720
     color: Colors.gray_500
-    
+
```
jwise, line 18: please avoid noise whitespace changes

bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
```diff
@@ -92,13 +92,19 @@ Rectangle {
...
+            if (resp.byteLength !== 0 && rv !== 0) { /* verify that install.sh exists and.. isn't blank? */
+                statusModel[statusModel.length - 1][0] = "success";
+                statusModel = statusModel;
+                secondaryStatusText = "";
+                X1PlusNative.system('mount -o remount,exec /userdata'); // Firmware R mounts it noeexec
+                X1PlusNative.system(`/userdata/x1plus/install.sh &`); // this will soon start communicating with us over DDS, hopefully
+            } else {
+                console.log("[x1p] failed to unpack x1p file");
+                dialogStack.replace("SelectX1pPage.qml"); //send user back to selection page
+            }
```
jwise, line 107: I think without a visible diagnostic, it is not obvious to me that just bouncing backwards is any more 'obvious' as to what is going on?  maybe the answer is just to push a message saying "X1P extraction failed.  The X1P file may be corrupt.  Replace the X1P file on the SD card and try again.'?

installer/install.py
```diff
@@ -222,19 +224,15 @@ def validate_sd():
...
-
 validate_sd()
 
 df = shutil.disk_usage("/mnt/sdcard")
 if df.free < demand_free_space:
     report_failure(f"The SD card does not have enough free space. Make sure that at least {demand_free_space//(1024*1024)} MB are available.")
 
 if not os.path.isfile("/sys/devices/virtual/misc/loop-control/dev"):
-    if os.system(f"insmod {installer_path}/loop.ko") != 0:
+   if os.system(f"insmod {installer_path}/loop.ko") != 0:
```
jwise, line 234: local Python style is 4 spaces, not 3 space indentation (and whitespace is significant in Python!)

installer/install.py
```diff
@@ -268,42 +266,70 @@ def debug_wifi():
 
 report_success()
 
-debug_wifi()
+os.system("/userdata/x1plus/debug_wifi.sh")
+
+def download_firmware(update_url, dest_path):
+    headers = {
+        'User-Agent': f"Python-urllib/3.{sys.version_info.minor}"
+    }
```
jwise, line 274: `urllib` is the wrong user agent here (and how we got ourselves into this problem to start with...).  How about `f"X1Plus/{cfw_version}"`?  Then we identify ourselves correctly.

installer/install.py
```diff
@@ -268,42 +266,70 @@ def debug_wifi():
...
+        except (requests.Timeout, requests.exceptions.HTTPError, requests.exceptions.RequestException) as e:
+            report_interim_progress(f"{e.__class__.__name__}: {e} ... retry {retry}")
+            time.sleep( pow( 2, retry ) if retry < 4 else 16 )
+            retry += 1
+        except Exception as e:
+            report_interim_progress(f"An error occurred: {e} ... retry {retry}")
+            time.sleep( pow( 2, retry ) if retry < 4 else 16 )
+            retry += 1
+        else:
+            success = True
```
jwise, line 294: This retries if the initial connection setup fails, but not if the download later fails (i.e., if we download fewer than `totlen` bytes), right?  Is this the desired behavior?

bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
jphannifan: There are several spots in the installer where it would be very nice to have a dialog prompt with a short description of the error and the dialog options are "Back" to the selection screen and "Reboot"

installer/install.py
jphannifan: This could be avoided for the most part by running `ping` before the download is attempted. If the ping fails, then don't download. This is also a good idea in general, because we don't really know what our margins for error are on timing for avoiding this race condition. It's not exactly a race condition but you know what I mean.

riptidewave93 (approved):

installer/install.py
```diff
@@ -268,42 +266,62 @@ def debug_wifi():
...
-basefw_squashfs_path = f"{boot_path}/images/{basefw_squashfs}"
-if not exists_with_md5(basefw_squashfs_path, basefw_squashfs_md5):
-    # maybe we can at least skip downloading it?
-    basefw_update_path = f"{boot_path}/firmware/{basefw_update}"
-    if not exists_with_md5(basefw_update_path, basefw_update_md5):
-        report_progress("Downloading base firmware")
-        if not ask_permission(f"Base firmware does not exist on the SD card: {basefw_update}<br><br>Download it from Bambu Lab servers?"):
-            report_failure(f"Base firmware {basefw_update} does not exist on the SD card.  Place it in {boot_path}/firmware and try again.")
+def download_firmware(update_url, dest_path):
+    if os.system("ping -c 3 -W 1000 8.8.8.8") != 0:
```
riptidewave93, line 271: Idea: maybe try pinging the actual hostname of the basefw update, so we test DNS and connectivity to the actual DL endpoint for what we are downloading. Not a blocker, just a suggestion.

installer/install.py
bani6809: there are other errors that can occur even if ping works. we can add totlen checking as well.

jphannifan: Right, the only thing this can serve as a breakpoint for is when a printer is able to ping google but not connect to Bambu's CDN links. 

After the issues we've been having, I would say the more breakpoints the better

installer/install.py
jphannifan: That's a good idea, but @bani6809 has a good point above too. If there's an issue with connecting to the CDN link, we can determine that by checking how many bytes have been downloaded. 

The goal of this is just to identify whether a user has a working connection but can't download from CDN links (which was Issue #184)

bani6809 (approved):
this PR survives my test which is:
1) turn off wifi AP
2) answer "yes" to prompt to download base firmware
3) turn on wifi AP
4) watch it fail several retries
5) watch it recover and eventually succeed

jwise (approved):
I trimmed the noise whitespace changes and made the installer complain properly if unpack fails.  This looks good now.  Thanks for the changes!

installer/install.py
```diff
@@ -222,9 +223,6 @@ def validate_sd():
     if not found_sdcard:
         report_failure(f"SD card was not mounted in the expected location. If the SD card is inserted, please file a bug.")
 
-def debug_wifi():
```
jphannifan, line 225: how did this happen?

PR #247, opened 2024-05-07
Update base firmware to 01.08.00.00.

### LINKED ISSUE #242: upgrade base firmware to 01.08.00.00
upgrade base firmware to 01.08.00.00

jwise:
Updates the base firmware to 01.08.00.00.  In so doing, we make two other changes required to make 01.08.00.00 work properly:

* When downloading in the installer, we identify ourselves as X1Plus, rather than urllib, because urllib is blocklisted on bblmw.com.  And anyway, we should do that.
* Because bblmw.com uses Let's Encrypt instead of Google SSL, it can break Qt's SSL implementation.  Remove an ancient expired Let's Encrypt certificate to fix that.  For bonus points, this also fixes a bug that is almost certainly present in OEM where avatars may not download properly since they were migrated to bblmw.com.

Fixes #242.

### DIFF
--- a/firmwares/Makefile
+++ b/firmwares/Makefile
@@ -20,6 +20,11 @@ define fetchimage =
 	../scripts/repack_update.py $(3) $(4)
 endef
 
+define fetchimage_fullurl =
+	wget -q -N -c "$(1)/$(2)"
+	../scripts/repack_update.py $(2) $(3)
+endef
+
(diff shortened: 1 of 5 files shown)


bani6809 (approved):
tested, works

jphannifan (commented):
works for me

PR #248, opened 2024-05-07
add back geftpserver and symlink to sftp-server from 01.07.00.00

bani6809:


jwise (approved):
if you have tested this I am happy with it

jphannifan (approved):

PR #249, opened 2024-05-07
fix: improvements to x1plusd settings

riptidewave93:
* Update settings with defaults, add boot flag file migrations
* Move away from flag files in all init scripts

CC @jwise this is the PR from last time, with OTA removed, and your requested changes added. OTA will stay in the old PR. :)


jphannifan (approved):
Looks good to me!

jwise (approved):
Looks good to me.  Thanks!

PR #250, opened 2024-05-08
Update language interposer for 01.08.00.00.

### LINKED ISSUE #229: memory map of bbl_screen languages changed again
```
LANG INTERPOSE: saw the first pass of en, the last QObject must have been the weird not-a-deviceManager
LANG INTERPOSE: here is the actual langmap init en for round 1
LANG INTERPOSE: DeviceManager's maps are probably ready, let's do it. langmap = 0xa6abd0e4, contents are now QMap(("de", "Deutsch")("en", "English")("es", "Español")("fr", "Français")("nl", "Nederlands")("ru", "Русский")("zh_CN", "中文"))
LANG INTERPOSE: found init for DeviceManager
LANG INTERPOSE: saw en for round 2 map initializer
LANG INTERPOSE: this bbl_screen does NOT look like the memory mapping we expect... not injecting languages into this version of bbl_screen
```

jwise:
The memory layout and call order for DeviceManager init changed somewhat in 01.08.00.00.  Fix the language interposer to deal with the new init sequence.

Fixes #229.

### DIFF
--- a/bbl_screen-patch/interpose.cpp
+++ b/bbl_screen-patch/interpose.cpp
@@ -776,16 +776,6 @@ SWIZZLE(void *, _ZN7QObjectC2EPS_, void *p1, void *p2)
 
 static QMap<QByteArray, QString> *langmap;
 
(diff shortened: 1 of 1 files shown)


jphannifan:
+1, nice work!

jphannifan (approved):
Looks like langmap just got moved in the memory map right? nice work

jwise:
Yeah, exactly.  A little bit of ghidra found the right answer.  And it was nice to know that the don't-crash detection failed gracefully!

PR #257, opened 2024-05-10
Adding debug_wifi.sh back

jphannifan:
debug_wifi.sh is no longer called from install.py

This might have been an unintended casualty when I was fixing merge conflicts for install.py in the recent PR, but regardless this was not the intended result. I've added it back and now it runs if the ping check fails.


riptidewave93 (approved):

bani6809 (approved):
👍

PR #259, opened 2024-05-11
QProcess class for subprocesses in QML

jphannifan:
Defines QProcess class in interpose.cpp to use in QML. 

Many of the methods are built into QProcess, so I have left the class very bare bones. We still have access to methods for setting environmental variables, working directory, etc. in addition to signals emitted with error codes and process status. 

Here is an example of using it to untar an archive:
```
import Process 1.0

extract.start("tar", ["-xvf", "cfw.x1p"])

 Process {
    id: extract
    property string output: ""
    onStarted: console.log("started")
    onFinished: console.log("Process finished with exit code:", exitCode, "and status:", exitStatus)
    onErrorOccurred: console.log("Error Ocuured: ", error)

    onReadyReadStandardOutput: {
        output = extract.readAll();
        console.log(output);
    }
}
```

This is for use anywhere in QML that requires efficient handling of shell commands and/or python. One idea I've already tested and have working on another branch is running ConsolePage.qml commands with this (although long term, I still want a real terminal emulator).

### DIFF
--- a/bbl_screen-patch/interpose.cpp
+++ b/bbl_screen-patch/interpose.cpp
@@ -24,6 +24,8 @@
 #include <stdlib.h>
 #include <QtCore/QObject>
 #include <QtCore/QSettings>
+#include <QtCore/QProcess>
+#include <QtCore/QVariantList>
 #include <QtQml/qqml.h>
 #include <QtQml/qjsengine.h>
 #include <QtQml/qjsvalue.h>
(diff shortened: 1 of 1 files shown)


riptidewave93 (approved):

jwise (approved):
This is clearly the right answer.  Nice one!  Only suggestion is that it might be nice to call this like `X1PlusProcess` to avoid polluting the global namespace.

PR #260, opened 2024-05-12
kexec_ui OTA autoinstall

jphannifan:
This branch adds logic and dialog to kexec_ui and /kexec/check_kexec to make it possible to "auto install" an x1p file. Once you have set the OTA flag in settings.json, the next time you reboot your device it will start installation automatically.

How to test:
1) Install the artifact build from this PR
2) Find an x1p file you want to install and save the filename to `ota.filename`. You can do this via ssh by running the following command, replacing '1.1rc1.x1p' with your target x1p filename.
```
python3 /opt/x1plus-test-settings.py ota.filename=1.1rc1.x1p
```

3) Reboot or power cycle and there should be no further input necessary to complete the update.

Any failures the updater experiences still require manual reboots, but any potential issues with wifi can be avoided by ensuring that the base firmware the x1p requires is on the sd card.

### DIFF
--- a/bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
+++ b/bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
@@ -52,33 +52,18 @@ Rectangle {
                     statusList.positionViewAtEnd();
                 }
                 if (datum.prompt_yesno) {
-                    textConfirmTimer.start()
                     dialogStack.popupDialog("TextConfirm", {
                         name: "installer yesno",
-                        text: Qt.binding(function() {  return datum.prompt_yesno + qsTr(" Download beginning in %1 seconds.").arg(textConfirmTimer.timerCount)}),
-                        titles: [qsTr("Download"), qsTr("Cancel")],
-                        finished: Qt.binding(function() { if (!textConfirmTimer.running) return true}),
+                        text: datum.prompt_yesno,
+                        titles: [qsTr("Yes"), qsTr("No")],
                         onYes: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: true })); dialogStack.pop(); },
-                        onNo: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: false })); dialogStack.pop(); },
-                        onFinished: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: true })); dialogStack.pop(); }
+                        onNo: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: false })); dialogStack.pop(); }
                     });
                 }
             }
         }
     }
-    Timer {
-        id: textConfirmTimer
-        interval: 1000
-        repeat: true
-        running: false
-        property int timerCount:10
-        onTriggered: {
-            timerCount--;
-            if (timerCount <= -1){
-                textConfirmTimer.stop();
-            }
-        }
-    }
+    
     Timer {
         /* This is super hokey, to allow a paint after we say 'unpacking
          * firmware bundle' but before we go to sleep for a while doing it. 
--- a/bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
+++ b/bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
@@ -55,7 +55,7 @@ Rectangle {
                     textConfirmTimer.start()
                     dialogStack.popupDialog("TextConfirm", {
                         name: "installer yesno",
-                        text: datum.prompt_yesno + qsTr(" Download beginning in %1 seconds.").arg(textConfirmTimer.timerCount),
+                        text: Qt.binding(function() {  return datum.prompt_yesno + qsTr(" Download beginning in %1 seconds.").arg(textConfirmTimer.timerCount)}),
                         titles: [qsTr("Download"), qsTr("Cancel")],
                         finished: Qt.binding(function() { if (!textConfirmTimer.running) return true}),
                         onYes: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: true })); dialogStack.pop(); },
--- a/bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
+++ b/bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
@@ -52,18 +52,33 @@ Rectangle {
                     statusList.positionViewAtEnd();
                 }
                 if (datum.prompt_yesno) {
+                    textConfirmTimer.start()
                     dialogStack.popupDialog("TextConfirm", {
                         name: "installer yesno",
-                        text: datum.prompt_yesno,
-                        titles: [qsTr("Yes"), qsTr("No")],
+                        text: datum.prompt_yesno + qsTr(" Download beginning in %1 seconds.").arg(textConfirmTimer.timerCount),
+                        titles: [qsTr("Download"), qsTr("Cancel")],
+                        finished: Qt.binding(function() { if (!textConfirmTimer.running) return true}),
                         onYes: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: true })); dialogStack.pop(); },
-                        onNo: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: false })); dialogStack.pop(); }
+                        onNo: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: false })); dialogStack.pop(); },
+                        onFinished: function() { DdsListener.publishJson("device/request/upgrade", JSON.stringify({ command: 'x1plus', yesno: true })); dialogStack.pop(); }
                     });
                 }
             }
         }
     }
-    
+    Timer {
+        id: textConfirmTimer
+        interval: 1000
+        repeat: true
+        running: false
+        property int timerCount:10
+        onTriggered: {
+            timerCount--;
+            if (timerCount <= -1){
+                textConfirmTimer.stop();
+            }
+        }
+    }
     Timer {
         /* This is super hokey, to allow a paint after we say 'unpacking
          * firmware bundle' but before we go to sleep for a while doing it. 
--- a/bbl_screen-patch/kexec_ui/printerui/qml/Screen.qml
+++ b/bbl_screen-patch/kexec_ui/printerui/qml/Screen.qml
@@ -35,11 +35,12 @@ Rectangle {
     
     Component.onCompleted: {
         var KEXEC_LAUNCH_INSTALLER = X1PlusNative.getenv("KEXEC_LAUNCH_INSTALLER");
-        var X1P_OTA = X1PlusNative.getenv("X1P_OTA"); //string, x1p filename
-        if (KEXEC_LAUNCH_INSTALLER != "")
-            dialogStack.popupDialog("KexecDialog", {startupMode:"installerSelect",countdown: 0}); //startupMode=installerSelect  ---> loads SelectX1pPage.qml
-        else
-            dialogStack.popupDialog("KexecDialog", {x1pName:X1P_OTA,  countdown: 10}); 
+        var X1P_OTA = X1PlusNative.getenv("X1P_OTA");
+        if (!X1P_OTA) X1P_OTA = "";
+        if (KEXEC_LAUNCH_INSTALLER != "") //stage 1 of the installer
+            dialogStack.popupDialog("KexecDialog", { stage1: KEXEC_LAUNCH_INSTALLER, countdown: 0});
+        else //normal boot and OTA updates
+            dialogStack.popupDialog("KexecDialog", {x1pName:X1P_OTA, countdown: 10}); 
     }
     
     Component.onDestruction: {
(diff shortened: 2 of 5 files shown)


bbl_screen-patch/kexec_ui/printerui/qml/Screen.qml
```diff
@@ -34,10 +34,15 @@ Rectangle {
     color: Colors.gray_500
     
     Component.onCompleted: {
-        if (X1PlusNative.getenv("KEXEC_LAUNCH_INSTALLER") != "")
-            dialogStack.push("SelectX1pPage.qml", { noBackButton: true } );
-        else
-            dialogStack.popupDialog("KexecDialog", {});
+        /* Boot flags set in check_kexec */
+        var KEXEC_LAUNCH_INSTALLER = X1PlusNative.getenv("KEXEC_LAUNCH_INSTALLER") || "";
+        var BOOT_DIALOG_TIMER = X1PlusNative.getenv("BOOT_DIALOG_TIMER") || 10;
```
jwise, line 39: I think this will not work right: getenv returns a string.

bbl_screen-patch/kexec_ui/printerui/qml/Screen.qml
```diff
@@ -34,10 +34,15 @@ Rectangle {
...
-            dialogStack.push("SelectX1pPage.qml", { noBackButton: true } );
-        else
-            dialogStack.popupDialog("KexecDialog", {});
+        /* Boot flags set in check_kexec */
+        var KEXEC_LAUNCH_INSTALLER = X1PlusNative.getenv("KEXEC_LAUNCH_INSTALLER") || "";
+        var BOOT_DIALOG_TIMER = X1PlusNative.getenv("BOOT_DIALOG_TIMER") || 10;        
+        var X1P_OTA = X1PlusNative.getenv("X1P_OTA") || "";
+
+        if (KEXEC_LAUNCH_INSTALLER != "") //first stage installer is running, skip dialog
+            dialogStack.popupDialog("KexecDialog", { stage1: KEXEC_LAUNCH_INSTALLER, countdown: 0});
```
jwise, line 43: if this skips the dialog why is it launching the dialog rather than just pushing SelectX1pPage?

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
```diff
@@ -16,34 +18,81 @@ Item {
...
             visible: countdown > 0
         }
+        DialogButtonItem {
+            name: "cancel"
+            title: startupStrings[startupMode].buttons["cancel"].text
+            isDefault: defaultButton == 2
+            onClicked: startupStrings[startupMode].buttons["cancel"].action()
+            visible: hasSdCard && countdown > 0 && startupStrings[startupMode].buttons["cancel"].text.length > 0
+        }
+    }
```
jwise, line 78: I feel like all this logic is trying to be three (?) different dialogs when it is in fact not three different dialogs...?

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
```diff
@@ -70,10 +120,9 @@ Item {
         font: Fonts.body_36
         color: Colors.gray_100
         wrapMode: Text.Wrap
-        text: hasSdCard ? qsTr("Bootable SD card detected.")
-                        : qsTr("No bootable SD card detected.")
+        text: hasSdCard ? startupStrings[startupMode].title : qsTr("No SD Card detected.")
```
jwise, line 123: ??? this is not what this flag means

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
```diff
@@ -84,10 +133,6 @@ Item {
         font: Fonts.body_32
         color: Colors.gray_200
         wrapMode: Text.Wrap
-        text: hasSdCard 
-            ? (countdown > 0) 
-                ? qsTr("Your printer will automatically boot from the SD card in %1 seconds.").arg(countdown)
-                : qsTr("Your printer is rebooting into the OS on the inserted SD card.")
-            : qsTr("Insert a bootable SD card and restart the printer, or use the startup options menu to repair your X1Plus installation.")
+        text: hasSdCard ? startupStrings[startupMode].subtitle() : qsTr("Please insert an SD card to continue.")
```
jwise, line 136: did you test this?  inserting a SD card will not clear this flag, as written, and this can happen if there is a SD card but it does not contain a boot.conf.

internal-fs/opt/kexec/check_kexec
```diff
@@ -26,6 +26,24 @@ elif [ -f /mnt/sdcard/x1plus/printers/${BBL_SN}/quick-boot ]; then
...
+    if [ "${OTA_FILENAME}" != "null" ]; then
+        if [ -f /mnt/sdcard/${OTA_FILENAME} ]; then
+            export X1P_OTA=${OTA_FILENAME}
+            QUICKBOOT_FLAG="false" # Back to false, so we make it to installer if we are re-installing a matching named build
+        fi
+        # Always clear the flag no matter what if it wasn't null, we have to do this ourselves as we don't have dbus
+        # for x1plusd as we are NOT in x1plus yet!
+        TMPSETTINGS=$(mktemp)
+        /opt/kexec/jq ".\"ota.filename\" = null" /mnt/sdcard/x1plus/printers/${BBL_SN}/settings.json > "${TMPSETTINGS}"
+        mv "${TMPSETTINGS}" /mnt/sdcard/x1plus/printers/${BBL_SN}/settings.json 2> /dev/null # suppress permissions warning
```
jwise, line 42: this cross filesystem move does not update things atomically the way you want it to: it unlinks the settings.json before moving the new one in place.  write to a temporary file in the SD card, *then* sync, *then* mv it in place over the top of the old one, *then* sync again

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
jphannifan: > I feel like all this logic is trying to be three (?) different dialogs when it is in fact not three different dialogs...?

There are just two sets of dialog. No dialog to display if the first stage installer is running. This is overly complicated though, you are right about that.

bbl_screen-patch/kexec_ui/printerui/qml/Screen.qml
jphannifan: Good question. I will find a way to simplify this

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
jphannifan: I will change it back to its original caption. New strings here = more translations so there was no point in changing to begin with.

internal-fs/opt/kexec/check_kexec
riptidewave93: I think there's a misunderstanding; jq isn't removing the file, it's modifying the output of it and saving it to tmp. We then move it ontop of the existing settings.json and syncing. Because of this there's never a moment where the file doesn't exist; this is the same logic as emmc_dump uses to clear it's flag.

bani6809:
kexec_ui appears to have lost its boot countdown display.

internal-fs/opt/kexec/check_kexec
jwise: Arguably, emmc_dump is wrong then, too :-)

The problem is not in the `jq` but in the `mv`.  In this case, the syscalls that `mv` executes, morally, are as follows:

```
open(TMPSETTINGS, O_RDONLY) -> fd_in
open("/mnt/sdcard/.../settings.json", O_WRONLY | O_CREAT | O_TRUNC) -> fd_out
read(fd_in, ...)
write(fd_out, ...)
close(fd_in)
close(fd_out)
```

This is not atomic: after the `O_TRUNC` open, if power is lost or the system crashes, settings.json ends up zero bytes.  What you want is something that uses the `rename` syscall, which, unfortunately, does not work cross-filesystem; from `rename`'s man page: "If newpath already exists, it will be atomically replaced, so that there is no point at which another process attempting to access newpath will find it missing."  (We do this already in [interpose.cpp](https://github.com/X1Plus/X1Plus/blob/main/bbl_screen-patch/interpose.cpp#L256-L259).)

The good news is that there is a tool that uses the `rename` syscall!  Indeed, that tool is ... `mv`.  The catch is that the source and destination need to be on the same filesystem.

So what you probably want to do is something like:

```
jq ... /mnt/sdcard/.../settings.json > /mnt/sdcard/.../settings.json.tmp
sync # make sure settings.json.tmp is fully committed to disk before we start writing the directory entry
mv /mnt/sdcard/.../settings.json.tmp /mnt/sdcard/.../settings.json
sync # make sure the directory entry is committed to disk
```

bani6809:
manual install x1p now has an auto countdown reboot timer. i do not believe this is correct.

jphannifan:
> kexec_ui appears to have lost its boot countdown display.

I fixed this one here: https://github.com/X1Plus/X1Plus/pull/260/commits/ad0aa41b6177b4f348bc7369b9351264727c54ea#diff-72e6dab8184dc884e7cc7768a2ad11e823c81e6e3863f7ea92b64b3deebee893R33

bbl_screen-patch/kexec_ui/printerui/qml/InstallingPage.qml
```diff
@@ -52,18 +52,33 @@ Rectangle {
...
+        repeat: true
+        running: false
+        property int timerCount:10
+        onTriggered: {
+            timerCount--;
+            if (timerCount <= -1){
+                textConfirmTimer.stop();
+            }
+        }
+    }
```
jwise, line 81: Hm, wait, I disagree that we should automatically download.  The original goal of this prompt is to avoid downloading things in the case of users who are paranoid about downloading things from Bambu.  I think that the new x1plusd.ota changes should mean that this dialog more or less shouldn't pop up any more, except for during initial install -- so maybe we do not want a countdown here after all?

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
```diff
@@ -16,34 +17,78 @@ Item {
         return xhr.status == 200;
     }())
     property var countdown: 10
+    property string currentMode: (x1pName !== "") ? "ota" : "default"
```
jwise, line 20: I think I would probably spell this as a bool rather than as a string, given that it has only two settings, though I guess using this as a key for startupStrings makes some sense

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
```diff
@@ -16,34 +17,78 @@ Item {
...
+        "ota": {
+            "buttons": {/* auto install x1p Boot: 10 sec dialog then tell InstallingPage.qml to start installing x1p */
+                "yes_confirm": {"text": qsTr("Install"), "action": function() { dialogStack.replace("../InstallingPage.qml", {x1pName: x1pName}) }},
+                "no": {"text": qsTr("Startup options"), "action": function() { dialogStack.replace("../BootOptionsPage.qml", {hasSdCard: hasSdCard}) }},
+                "cancel": {"text": qsTr("Boot X1Plus"), "action": function() { X1PlusNative.system("/opt/kexec/boot"); }}
+            },
+            "title": qsTr("Update ready to install!"),
+            "subtitle": function() {
+                if (countdown > 0)
+                    return qsTr("Automatically installing .x1p file '%1' in %2 seconds. To cancel or skip installation, select 'Startup options' or 'Boot X1Plus'.").arg(x1pName).arg(countdown);
```
jwise, line 47: I will tweak the wording of this.  also I wonder if we should just have 'install | startup options'

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
```diff
@@ -16,34 +17,78 @@ Item {
...
+            isDefault: defaultButton == 2
+            onClicked: startupStrings[currentMode].buttons["cancel"].action()
+            visible: hasSdCard && countdown > 0 && startupStrings[currentMode].buttons["cancel"].text.length > 0
+        }
+    }
+
+
+    Component.onCompleted: {
+
     }
```
jwise, line 81: trim null block

bbl_screen-patch/kexec_ui/printerui/qml/dialog/KexecDialog.qml
```diff
@@ -70,10 +116,9 @@ Item {
         font: Fonts.body_36
         color: Colors.gray_100
         wrapMode: Text.Wrap
-        text: hasSdCard ? qsTr("Bootable SD card detected.")
-                        : qsTr("No bootable SD card detected.")
+        text: hasSdCard ? startupStrings[currentMode].title : qsTr("No bootable SD Card detected.")
```
jwise, line 119: keep original capitalization to avoid translation churn

bbl_screen-patch/kexec_ui/printerui/qml/dialog/Success.qml
```diff
@@ -5,37 +5,60 @@ import "qrc:/uibase/qml/widgets"
...
+            if (countdown <= 0) {
+                doReboot();
+            }
+        }
+    }
+    
+   property var buttons: SimpleItemModel {
+        DialogButtonItem {
+            name: "reboot"
+            title: countdown > 0 ? qsTr("Reboot in ") + countdown + qsTr(" seconds") : qsTr("Rebooting..")
```
jwise, line 45: translation should use .arg to interpolate

jwise (commented):
This makes sense to me.  Let me take a pass on this, I will see if I can make a few changes to clean a few of these up?  Generally this seems good though.  Sorry for the delay on the review!

jwise (approved):
woo I have just taken an OTA without touching the printer.  this rules

PR #262, opened 2024-05-14
x1plusd.settings: avoid blowing up if the settings file is corrupt on load

jwise:
Previously, if the settings.json was corrupt, then x1plusd would simply fail to start.  That is certainly not ideal.  Avoid this.


jphannifan (approved):
This works for me. 

Before fix if I created a blank settings.json file, x1plusd.py it crash

After fix I make blank settings json file, x1plusd no crash. is fixed!

riptidewave93 (approved):

PR #265, opened 2024-05-15
feat: add OTA engine

jwise:
TODO:
* add support to download an x1p, and to download Bambu fw (and make that generate status events)
  * make sure we have enough storage to download the x1p?
* either teach the ota.json generator to embed the buildTimestamp, or use the OTA date instead
* wire it up to kexec_ui autoinstall
* maybe write a CLI tool for this?  or should that be in its own PR?

I was testing it on my laptop (have not tested on printer HW yet) with various combinations of:

`joshua@anima:~/bambu/x1plus/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd$ PYTHONPATH=../../.. python3 -m x1plus.services.x1plusd`

`$ python3 ./x1plus-test-settings.py ota.enabled=True`

`$ dbus-send --session --print-reply --dest=x1plus.x1plusd /x1plus/ota x1plus.ota.CheckNow string:'{}'`

`$ dbus-send --session --print-reply --dest=x1plus.x1plusd /x1plus/ota x1plus.ota.GetStatus string:'{}'`

`$ python3 ./x1plus-test-settings.py ota.enabled=None`


jwise:
chatted with @riptidewave93 -- we think this is a good v1, and the rest of the items in this list can go in after this PR.  I haven't tested on HW but if someone else has, feel free to +1, and then we can merge it

riptidewave93 (approved):

PR #266, opened 2024-05-15
Print speed adjustment + documentation of OEM speed profiles

jphannifan:
The question I set out to ask initially was, "how do I keep extrusion profiles constant when adjusting speeds mid-print?". In answering this question I created some useful documentation which can become part of the X1Plus wiki [here](https://github.com/jphannifan/x1plus-testing/blob/main/BL-speed-adjust.md). 

1) Documentation and creation of gcode function:
A quick google search confirms that we are providing the first documentation of Bambu's print speed gcode. The parameters and gcode sequences for each speed level can only be found by monitoring bbl_screen or other services, or viewing log data. See the readme linked above

2) Gcode function:
X1Plus.GcodeGenerator.printSpeed() approved in [this PR](https://github.com/X1Plus/X1Plus/blob/d7038787ac6f1d69373a80e792b69d40d5673874/bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js#L521)
```
gcode = X1Plus.GcodeGenerator.printSpeed(100)
X1Plus.sendGcode(gcode)
```

3) QML object: SpeedAdjust.qml
With minor modifications to AdjustPage.qml, this provides an alternative interface for speed adjustments without removing or replacing OEM speed adjust UI. SpeedAdjust.qml allows you to adjust your print speed from 30 to 180%. This is a Dial type that uses custom canvas drawings for annotation. It started out as a slider, and it was hard to control so now it's a dial. 

I have had 3 users test this feature from a branch over the past month, and have no issues to report, and I personally have a couple hundred print hours of testing the final Gcode function now. I'm happy to share photos of prints but the Gcode was approved about a month ago now. This PR is for documentation and the UI 

Big thanks to [WolfwithSword](https://github.com/WolfwithSword) for helping me test this. He added a speed adjustment slider to his HA workflow last month which uses the same trendlines as the X1Plus function, so anyone with an X1, A1, or P1 can use this feature without firmware mods.


riptidewave93 (approved):

jwise:
Oop, I didn't realize that this went from a draft to a full PR -- I saw it only when it was a draft!  I'll take a look tomorrow evening.  I see a few code cleanups that I want to make but broadly this looks pretty good!

jphannifan:
> Oop, I didn't realize that this went from a draft to a full PR -- I saw it only when it was a draft! I'll take a look tomorrow evening. I see a few code cleanups that I want to make but broadly this looks pretty good!

I had a bug to fix after I published it, so I just now got around to assigning reviewers. I do think a slider could work better here, because the dial definitely takes some getting used to. There's no rush on this if you want to take a crack at a slider (or anything else here)

jwise:
OK so I took a look at this tonight.  Here's kind of my arc of thoughts about it:

First off, the split between `AdjustPage` and `SpeedAdjust.qml` is really nice.  I wish Bambu had done that for everything, actually.  It makes it so much cleaner and it makes it obvious where to look for the speed dial.  Nice work!

A handful of things I want to think about / take a crack at changing:

- [ ] I think the `speedStr` thing can be replaced by `PrintManager.currentTask.printSpeedModeStrings` and then we get translations?
- [ ] I am not so sure about this comment: `//EventTrack is what results in OEM speed gcode being published, so we want to avoid that here`.  Is that really true?  I thought `EventTrack` was basically UI / UX tracking and logging.  I thought the change happens in this? `PrintManager.currentTask.printSpeedMode = speedMode`
- [ ] One thing I am thinking about is that it would be nice for the `speedPad` to be wider when the `SpeedAdjust` is active.  We have a bunch of screen real estate and it would be nice to use it, especially given that we have a bunch of flavor text that we want to render.  Being able to swipe all the way from the left to the right would give quite a bit of finger resolution too.  I think I know how to make that happen at least.
- [ ] I want to think about a more intuitive switch mechanism than `<` and `>`.  Maybe a kebab menu option?  I spent most of a shower brainstorming graphics for this and kind of came up empty handed.
- [ ] I am still kind of concerned that "100%" actually is an acceleration of +2% and a feed rate of +5%
- [ ] I am not sure I understand some of the UI from looking at it.  For instance, in the following screenshot, "100%" appears in 4 places and I am not sure what it means in each place.
image
- [ ] The behavior seems strange and nonintuitive in emulation, but that could be because bbl_screen is getting confused in emulation.  I will try it on hardware next.  I also confess I do not understand what causes text to appear at left vs. a number.

Some of these I think I can take care of sort of brainlessly (UI tweaks, variable width).  Others of these I am going to have study what you did in more detail :-)

jphannifan:
There are a couple things that I should clarify, because it is not possible to "change speeds" in emulation. 
1) The speed displayed in the UI (right above the fan control panel) is a value calculated from the parameter of the Gcode command `M73.2 R`, rather than just an value from a hardcoded list. 
2) You can observe this directly by publishing `M73.2 R{100/speed}` to the device. You will notice that the speed percentage updates on the touchscreen, the slicer, and the app simultaneously. This behavior is the same on OEM too
3) The estimated time remaining is calculated from the same parameter (not ideal but we'll manage)

And to respond to your other comments:
- Those labels you're referring to are redundant, and they won't make the final cut. The label on the left is showing the current speed and the one on the right shows the target speed. I have no problem with removing both of these, and the current speed can be displayed somewhere on the dial. 
- The feed rate and acceleration differences are not ideal but I haven't found a good solution yet. simply applying an offset is an option, but that seems counterproductive to me.

(every time I click one of the checkboxes in your post, it counts as a revision to your message. oops)

jphannifan:
Added time estimate (only displayed when a print job is active)
image

riptidewave93:
@jphannifan any chance you could do a rebase instead of a squash/merge commit? your PR now reflects 659 changed files which makes it near impossible to review your PR.

jphannifan:
> @jphannifan any chance you could do a rebase instead of a squash/merge commit? your PR now reflects 659 changed files which makes it near impossible to review your PR.

ah damn, that was pretty dumb. Thanks for letting me know!

PR #267, opened 2024-05-15
feat: breakout and improve ota.json generation

riptidewave93:
* Make our ota.json generation be the final step, so we can include the file's MD5 for verification
* Also make sure the base FW info is attached, so our OTA engine can pull this down before we get to bbl_screen
* Also bring back gitignore for interpose.moc.h cuz it got lost somehow. :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 __pycache__
 localconfig.mk
 installer/info.json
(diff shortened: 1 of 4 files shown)


jwise (approved):
LGTM.  And we can also retroactively run mkotajson.py on 1.1 to fix up the ota.json in the field.

PR #270, opened 2024-05-18
make 'Discord' a hyperlink based on URL in the wiki

AskAlice:
ctrl+f user experience is important


jwise (approved):
Indeed! Thanks!

jwise:
I will rambo this change in when I get back to a keyboard because CI does not work on 3p pull requests. Ugh I wish I had a plan for how to deal with that.

PR #271, opened 2024-05-18
use new x1plus.services naming convention

bani6809:


jphannifan (approved):
We should be able to combine all 4 of these `x1plus.services.*` into a single service with just one init.d script. It sounds like we're headed that direction anyway by combining sd_watchdog and gpiokeys with x1plusd. 

I'm almost done configuring syslog-ng and afterwards we won't need syslog-shim at all anymore

riptidewave93:
> We should be able to combine all 4 of these `x1plus.services.*` into a single service with just one init.d script. It sounds like we're headed that direction anyway by combining sd_watchdog and gpiokeys with x1plusd.

This service is a C binary tho, and not python. This may need to stay outside of x1plusd for that reason, but no big deal. However, similar service naming for our stuff is a good idea so this PR is a +1 from me.

riptidewave93 (approved):

PR #272, opened 2024-05-19
x1plus/DBus: make sure to generate valid DBus messages even if passed no args

jwise:
otherwise, bbl_screen would choke on the initial GetSettings() call (it would pass 'undefined' up to DBus, rather than 'null'; x1plusd would fail to parse it, and return an error; and then bbl_screen would choke on the returned error code and fail to continue executing


riptidewave93 (approved):

jwise:
included in #273

PR #277, opened 2024-05-22
Interactive ConsolePage

jphannifan:
1) Shell console using QProcess to run subprocesses from QML and collect their stdout/stderr via signals. This means that it is now possible to run continuous processes and scripts that require user interaction.

2) Interactive Gcode console. This uses QProcess to run a helper script I wrote in Python that starts a thread in which syslog is tailed to look for gcode output, and meanwhile publishing Gcode commands entered by the user. We can get the "output" of a Gcode command by tailing syslog and extracting the text wrapped with these lines:
```
[MCU][BMC]owner=1 lock! [Free]
[MCU][BMC] .... 
[MCU][BMC]owner=1 unlock!
```
3) The command shortcuts are now loaded from JSON files (/usr/etc/macros/shell_commands.json and gcode_commands.json). So the command shortcuts can be customized now. 

4) Tooltips updated

Note: I still think we should replace the shell console with a QTermWidget in a future PR, so I want to avoid any significant changes to the shell console aside from bug fixes. 

Live tests:

https://github.com/X1Plus/X1Plus/assets/149451641/07f19aeb-dbec-4033-b762-57f3280806ca

https://github.com/X1Plus/X1Plus/assets/149451641/eaddb023-4906-4f42-b498-00657339e30c


jphannifan:
I'm taking this PR in a new direction(s). Instead of 1 PR over everything, I want to do the following:

1) PR for an x1plusd gcode generator
- port gcodegenerator.js to Python
- gcode command validation and sanitizing
- CLI for publishing commands

2) PR for a refactored gcode terminal/UI - 
- I want a new input method that combines a command selection screen with a numberpad to eliminate the virtual keyboard 
- log file tailing feature will be an optional debugging feature

3) Terminal emulator refactor
- Either replace it with an QTermWidget or use the work I've already done and just use QProcess

4) Gcode macros
- Bring back Gcode macros as a gpiokeys option

PR #280, opened 2024-05-24
Changes to Readme.

ima-hima:
I had some trouble making the .x1p file from the Dockerfile with the given instructions, so I made them way more detailed for the next guy like me who shows up and hasn't used Docker in a while.

Note that I left out `` `id -u` `` from all the docker commands. On the Mac that was causing problems, because it gives a different value than `` `id -u` `` does on Linux, and it didn't seem necessary, anyway. I did not test these commands on Linux or Windows, although since it's in a Docker container it _should_ be platform-agnostic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,20 +76,26 @@ $ docker exec -v `pwd`:/work x1plusmake sh -c 'ls /work | grep .x1p'
 $ <somefilename>.x1p
 ```
 
-Copy that out of the Docker container 
+Copy that out of the Docker container. 
 
 ```
-$ docker cp x1plusmake:/work/<somefilename>.x1p .
+$ docker cp x1plusmake:/work/<somefilename>.x1p
 ```
 
-And `scp` it to the printer. NEVER remove an SD card from a live X1Plus system, 
-you goofball!
+And `scp` it to the printer (never remove an SD card from a live X1Plus system, 
+you goofball).
 
 ```
-scp <somefilename>.x1p root@<printer’s IP>/sdcard
+scp <somefilename>.x1p root@<printer’s IP>:/sdcard
 ```
 
-Reboot your printer, and install it from the menu.
+Reboot your printer, and install X1Plus from the menu.
+
+Finally, don't forget to stop your Docker container!
+
+```
+docker stop x1plusmake
+```
 
 If you're going to make changes to any of the UI files, you should really
 read the rest of this document...  otherwise you might be in for a nasty
--- a/README.md
+++ b/README.md
@@ -29,13 +29,38 @@ container.  If you're adventurous, you can probably build X1Plus on any old
 Linux machine (I do), but if you do that and it breaks, we really don't want
 to hear about it, so you really should just build it in Docker.  You'll need
 the filesystem decryption key from a live printer (running either X1Plus or
-the Official Rootable Firmware) in order to build X1Plus; try something
-like:
+the Official Rootable Firmware) in order to build X1Plus
+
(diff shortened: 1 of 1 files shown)


jphannifan:
I tested it on Windows a while back and as long as you install and run WSL before cloning the repo and building the docker image, it will work.

So I think we could include this somewhere as a note to Windows users:

```
wsl --install
wsl -d Ubuntu
```

jphannifan:
@ima-hima Thanks for writing out the setup instructions by the way! This will help a lot of people that may have had trouble building the docker image.

scooterthedev:
> Note that I left out `` `id -u` `` from all the docker commands. On the Mac that was causing problems, because it gives a different value than `` `id -u` `` does on Linux, and it didn't seem necessary, anyway.

Funny thing is that Docker isn't really that platform-agnostic (at least from my experience). The command docker exec on WSL does not have the option of -v, and unless we run the other commands provided from Joshua (docker run), none of the commands will work, and it will refuse to build. It may be cleaner, and simpler to have the README formatted like this, but for Linux and WSL only, and have a dedicated page we can redirect Mac users too (maybe within the wiki), because at this moment, none of those commands will work for Linux or WSL (not sure about mac, as I am highly anti-apple lol). @jphannifan are you able to review these new commands from @ima-hima to make sure they work on a Mac?

ima-hima:
> > Note that I left out `` `id -u` `` from all the docker commands. On the Mac that was causing problems, because it gives a different value than `` `id -u` `` does on Linux, and it didn't seem necessary, anyway.
> 
> Funny thing is that Docker isn't really that platform-agnostic (at least from my experience). The command docker exec on WSL does not have the option of -v, and unless we run the other commands provided from Joshua (docker run), none of the commands will work, and it will refuse to build. It may be cleaner, and simpler to have the README formatted like this, but for Linux and WSL only, and have a dedicated page we can redirect Mac users too (maybe within the wiki), because at this moment, none of those commands will work for Linux or WSL (not sure about mac, as I am highly anti-apple lol). @jphannifan are you able to review these new commands from @ima-hima to make sure they work on a Mac?

Gah. That's so annoying. To be clear about what was happening, `id -u` on the Mac returns 501, at least if the user is the first registered user on the computer. That id increments for each new user added. When I attempted this on Ubuntu it returned 1000. However, my Ubuntu runs in a VM, so Docker doesn't actually work, as it needs KVM, which isn't available to me. Hard-coding 1000 did work on the Mac, but so did leaving out `-u` entirely. 

Since I'm complaining, also of note is that `gcc-multilib` doesn't seem to be available for ARM, so this won't compile at all on modern Apple hardware, and probably (due to the KVM issue above I wasn't able to actually test it) not on ARM Ubuntu, either.

ima-hima:
> @ima-hima Thanks for writing out the setup instructions by the way! This will help a lot of people that may have had trouble building the docker image.

You are so welcome!

jphannifan:
> Since I'm complaining, also of note is that `gcc-multilib` doesn't seem to be available for ARM, so this won't compile at all on modern Apple hardware, and probably (due to the KVM issue above I wasn't able to actually test it) not on ARM Ubuntu, either.

You should be able to use Rosetta to run x86_64 binaries on any modern Apple device. I see it was recently added to Docker and is no longer considered an experimental feature: https://www.docker.com/blog/docker-desktop-4-25/

I installed Docker Desktop a few months ago just to enable this setting, but I use Docker's CLI for everything else. The only other setup that may be required is installing Rosetta 
`softwareupdate --install-rosetta`

PR #288, opened 2024-06-08
try way harder on mounting sd

bani6809:


riptidewave93:
@bani6809 can we get some context on this? There's no description as to why were doing this change, and no linked GitHub issue. We need to be mindful how long we wait as flaky SDCards, while being slow to enumerate and mount, can cause other issues as well.

bani6809:
> @bani6809 can we get some context on this? There's no description as to why were doing this change, and no linked GitHub issue. We need to be mindful how long we wait as flaky SDCards, while being slow to enumerate and mount, can cause other issues as well.

i got this several times in a row on a samsung pro endurance. i'm pretty sure there are random timing conditions in the kernel that will cause a mount to fail that are not the fault of the card.

sd fail

RaySajuuk:
> > @bani6809 can we get some context on this? There's no description as to why were doing this change, and no linked GitHub issue. We need to be mindful how long we wait as flaky SDCards, while being slow to enumerate and mount, can cause other issues as well.
> 
> 
> 
> i got this several times in a row on a samsung pro endurance. i'm pretty sure there are random timing conditions in the kernel that will cause a mount to fail that are not the fault of the card.
> 
> 
> 
> sd fail
> 
> 

If I understand correctly, the SD card is mounted before the boot loader display on screen.

Possibility of corrupted file system, physical nand, or SD slot having issues?

jwise:
Yeah, I dunno, I'm a little worried about this one -- especially if we can't reproduce it.  I think it is arguably the case that if the SD card comes up in a weird tuning state, then maybe we should just fail to boot after all, and force a reboot to try again?  At least, without getting a `dmesg` of the failing case, I feel pretty nervous about trying to limp along in this state...

bani6809:
> Yeah, I dunno, I'm a little worried about this one -- especially if we can't reproduce it. I think it is arguably the case that if the SD card comes up in a weird tuning state, then maybe we should just fail to boot after all, and force a reboot to try again? At least, without getting a `dmesg` of the failing case, I feel pretty nervous about trying to limp along in this state...

maybe related to what balika011 suspected WRT reinit card without powercycle? https://discord.com/channels/1041781310299062282/1188616216936468542/1220843438149341184

PR #290, opened 2024-06-08
Add handshake with Polar Cloud

ima-hima:
Caveats:
1. Added Crypto and socketio libraries. These had some dependencies. I installed them using `pip install --target .images/site-packages` and added all of that (15 total packages) to the branch.
2. Had to add code to `__init__()` to launch Polar Cloud service.
3. There are a lot of todo comments about connecting to dbus to talk to interface.
4. Lack of interface
5. Am I dealing with `is_emulating()` logic okay? To this I need a login to Polar Cloud which I stored in a .env for now. This is obvs a horrible bandaid, but it's the only way I could figure out how to do this without adding the interface.
6. `x1plus-test-polar_cloud.py` could not be included until Polar Cloud is talking to dbus, which will be the next PR.


bbl_screen-patch/patches/printerui/qml/x1plus/Settings.js
```diff
@@ -63,7 +63,7 @@ function awaken() {
             _settingChanged(k, arg[k]);
         }
     });
-    _settings = X1Plus.DBus.proxyFunction("x1plus.x1plusd", "/x1plus/settings", "x1plus.settings", "GetSettings")() || {};
+    _settings = X1Plus.DBus.proxyFunction("x1plus.x1plusd", "/x1plus/settings", "x1plus.settings", "GetSettings")();
```
jphannifan, line 66: I think you may want to revert the change to this line because I don't see anything in `settings.py` that directly prevents `GetSettings` from returning null or undefined. When it does have a null or undefined value, I'm fairly sure that results in a JS type error on line 69.

ima-hima: I didn't actually edit this line. I haven't yet touched any of the Javascript files yet (that's coming in the next PR, when I add some Polar Cloud interface files), so I'm not sure what's happened. Is it possible that was in X1Plus's `main` at some point, and when I synced the fork it got pulled into my code? Or maybe I forked before the `or` was added in and git got confused on some merge or rebase?

The only files I changed were `.gitignore`, `polarcloud.py`, `__init__.py`, `utils.py`, and adding the Python libs.

In any case, yeah, I'll revert it.

jphannifan: I figured that may be the case. I don't know what specifically causes this, but it seems to happen when rebasing a branch. I've had this happen too with code I have never touched, especially with patch files. [This](https://github.com/X1Plus/X1Plus/pull/290/files#diff-3f631b44f07bfba7f0ec409ed4c31ab27b6b55baaf66d41139b102e16ee426c3L81) looks like it might be another instance of this issue.

ima-hima: Yeah, I did rebase at some point, which I rarely do, so I wasn't watching for this sort of thing. TIL.

I added those settings back.

jwise:
I think I am going to pull the Python packages into their own PR, probably to be merged after 2.0.  I saw a bunch of `.so` files in this commit, and those need to be built *very* carefully in order to work right on the printer.  In the mean time, I will go read through the rest of your code.  Thanks for putting up this PR to begin with!

Another thing I am realizing is that GitHub actions will *always* fail when faced with a pull request from another repository (because of the way GitHub secrets work).  We really should fix that.  But for now, if you are going to do development like this, maybe we should just kick the can down the road, and maybe we should just give you push permissions to a branch on this repository, so you can push to a branch like the rest of X1Plus development does.  @jphannifan @bani6809 @riptidewave93 What do you think about that?

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -18,6 +19,9 @@ async def main():
     settings = SettingsService(router=router)
     ota = OTAService(router=router, settings=settings)
     ssh = SSHService(settings=settings)
+    if settings.get("polar_cloud", False):
+        polar_cloud = PolarPrintService(settings=settings)
+        asyncio.create_task(polar_cloud.begin())
```
jwise, line 24: This is fine, but this means that if `polar_cloud` is turned off, then you will have to reboot the printer in order to get it to load `PolarPrintService` (or otherwise restart `x1plusd`.  For now, I am OK with this, out of an abundance of caution, but probably later we will want `PolarPrintService` to be loaded all the time and then just be careful about enabling or disabling itself.

jwise: Although actually, one more thought: if we *are* being careful about what we load or do not load here, maybe the `import` should be in this `if` block?  That way we can avoid the startup cost of loading all of `engineio` and its dependencies.  I'm not sure how much this actually costs in reality.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -18,6 +19,9 @@ async def main():
     settings = SettingsService(router=router)
     ota = OTAService(router=router, settings=settings)
     ssh = SSHService(settings=settings)
+    if settings.get("polar_cloud", False):
```
jwise, line 22: I probably would spell this key as `polar_cloud.enabled`

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+        # self.polar_settings.on("polarprint.enabled", self.sync_startstop())
+        # self.polar_settings.on("self.pin", self.set_pin())
+        self.socket = None
+        self.connected = False  # We might not need this, but here for now.
+        if is_emulating():
+            # I need to use actual account creds to connect, so we're using .env
+            # for testing, until there's an interface.
+            # dotenv isn't installed, so just open the .env file and parse it.
+            # This means that .env file must formatted correctly, with var names
+            # `username`, `pin`, and `server_url`.
```
jwise, line 48: This would be a good use of `x1plus.settings`!  In `is_emulating()` world, you can still use `x1plus settings`:

```
joshua@anima:~/bambu/x1plus/images/cfw/opt/x1plus/bin$ PYTHONPATH=../lib/python/ ./x1plus settings show vnc.enabled
true
```

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+from x1plus.utils import get_MAC, get_IP, serial_number, is_emulating
+
+logger = logging.getLogger(__name__)
+
+
+class PolarPrintService:
+    def __init__(self, settings):
+        self.polar_sn = 0
+        # Todo: VERY IMPORTANT!! The public and private keys MUST be moved to
+        # non-volatile memory before release.
```
jwise, line 27: How are these managed in the long term?  I assume they are generated on the printer, per-printer?  I could imagine either storing them in a file in `/mnt/sdcard/x1plus/printers/.../`, or maybe they should just live in the `x1plus.settings` store directly?

There is also very limited non-volatile configuration storage on eMMC (printer internal storage), but, until now, we have avoided trying to write to this.  If you think this needs to live on printer instead of on SD card, we probably should think long and hard about that, though.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+            # `username`, `pin`, and `server_url`.
+            import inspect
+
+            env_dir = os.path.dirname(inspect.getfile(inspect.currentframe()))
+            with open(os.path.join(env_dir, ".env")) as env:
+                for line in env:
+                    k, v = line.split("=")
+                    setattr(self, k, v.strip())
+
+    async def begin(self):
```
jwise, line 57: We've been calling the top-level threadlet for each module `task`, but I guess this is probably not actually going to be the top-level threadlet: this appears to connect once on printer startup, but does not attempt to repeatedly connect.  Ok for a WIP, I think.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+        self.socket.on("keyPair", self._on_keypair_response)
+        self.socket.on("helloResponse", self._on_hello_response)
+        self.socket.on("welcome", self._on_welcome)
+
+    async def _make_request(self, which_request, data: Dict[str, str]) -> None:
+        """Make a request to the server. Not currently in use. Delete?"""
+        if self.socket:
+            await self.socket.emit(which_request, data)
+        else:
+            logger.error(f"{which_request} failed. Socket is closed.")
```
jwise, line 78: handle the exception, not the rule, so that the success case is a straight line:
```suggestion
        if not self.socket:
            logger.error(f"{which_request} failed. Socket is closed.")
            return
        await self.socket.emit(which_request, data)
```

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+            self.public_key = response["public"]
+            self.private_key = response["private"]
+            # We have keys, but still need to register. First disconnect.
+            logger.info("_on_keypair_response success.\nDisconnecting.")
+            # Todo: I'm not creating a race condition with the next three fn calls, am I?
+            await self.socket.disconnect()
+            # After the next line it will send a `welcome`.
+            logger.info("Reconnecting.")
+            await self.socket.connect(self.server_url, transports=["websocket"])
+            await self._register()
```
jwise, line 160: is `connect` synchronous?  (well, as synchronous as `async`/`await` gets?)

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
```diff
@@ -16,3 +16,26 @@ def serial_number():
...
+        return "CC:BD:D3:00:3B:D5"
+    # Get the device id of the currently-active network device.
+    devices = subprocess.Popen(["ip", "link"], stdout=subprocess.PIPE)
+    # The active device _should_ be broadcasting.
+    active = subprocess.run(["grep", "BROADCAST"], stdin=devices.stdout, capture_output=True)
+    identifier = active.stdout.decode().split(":")[1].strip()
+    device = subprocess.Popen(["ip", "a", "s", identifier], stdout=subprocess.PIPE)
+    # Should be "ether" even if it's a wireless device.
+    ether = subprocess.run(["grep", "ether"], stdin=device.stdout, capture_output=True)
+    return ether.stdout.decode().strip().split(" ")[1]
```
jwise, line 33: let me think about how to do this without invoking `grep` at least :-) it might be the case that calling open `ip a s` by hand is the best we can do but we definitely can do better than calling `grep`

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
```diff
@@ -16,3 +16,26 @@ def serial_number():
...
+
+
+def get_IP() -> str:
+    """Return the IP address of the printer."""
+    if is_emulating():
+        return "192.168.2.113"
+    hostname = subprocess.run(["hostname", "-I"], capture_output=True)
```
jwise, line 40: this is going to get weird with a world in which `wlan0` can exist *and* `eth0` can exist, but we'll see how that ends up going.  (also, on the printer, `hostname -I` does not seem to work, sadly!)

jwise:
Broadly this seems good.  Since it's all gated out on load, I'm not super worried, but in general, it would be nice if settings were loaded in such a way that other X1Plus developers could test it too (i.e., with the x1plus settings mechanism to store server URL, with defaults, and with the x1plus settings mechanism to store username/pin/etc), rather than by hardcoding them in the constructors.

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
jphannifan: It's probably easiest if you check whether you need the address for eth0 or wlan0 ahead of time and then do
```
        with open(f'/sys/class/net/wlan0/address', 'r') as file:
            mac_address = file.read().strip()
        return mac_address
```

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
ima-hima: Yes, short term, my plan was to have to reboot each time. In fact, I think you'd have to rebuild the entire OS and ssh it across, wouldn't you? I _think_ @jwise we talked about this, but I _know_ I was doing it to avoid bloat, which is honestly something that matters to me.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
ima-hima: I actually tried that and was getting errors, so I was going to revisit it later. My next two planned PRs are to start talking to the dbus and interface, and start using X1Plus settings service. I can do that in either order. Let me know what you prefer.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
ima-hima: Hmm. Okay, I'll prioritize interacting with the settings service (and probably adding in some tests) over dbus.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
ima-hima: Yes, this is per-printer. I was figuring I'd put it on the SD card; it didn't occur to me that it could go in main printer memory.

I _think_ it wouldn't matter if it was per SD card, but let me talk to my colleagues and see what they think.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
ima-hima: Ugh, sorry. I realized that _after_ I shipped. It's actually the first thing I did after I did this PR, when I made PolarPrintService inherit from dbus, so it's planned for the next, or probably the subsequent, PR.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
ima-hima: Yes, `connect` is asynchronous when I use `socketio.AsyncClient()`.

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
ima-hima: Yeah, it turns out I need to use the same MAC whatever one is actually active, so I'll just pick wlan.

jwise:
Can you rebase onto dev/2.1?  Also, can you push to a branch on the X1Plus repository? (`polar/add-connect` would be a fine branch name.)  I think I gave you write access to the X1Plus repo.

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
ima-hima: > It's probably easiest if you check whether you need the address for eth0 or wlan0 ahead of time and then do
> 
> ```
>         with open(f'/sys/class/net/wlan0/address', 'r') as file:
>             mac_address = file.read().strip()
>         return mac_address
> ```

I don't know if I missed this comment and code or if it's newer than my last commit, but it's nicer than what I have, so I'll switch to this.

ima-hima:
> > Can you rebase onto dev/2.1? Also, can you push to a branch on the X1Plus repository? (`polar/add-connect` would be a fine branch name.) I think I gave you write access to the X1Plus repo.
> 
> Yes, I'll do both of these later today when I have a chance.

Okay, this is done. The rebase was a little messy, but it was I think all `.so` files, so I just used yours.

ima-hima:
Merged to dev/2.1 instead.

PR #292, opened 2024-06-10
Add EngineIO and PyFTDI python packages.

jwise:
#290 contains both Polar code and EngineIO/Pycryptodome; I split the Python bits into their own PR, and rebuild Pycryptodome to target arm32 rather than Darwin.  Then #290 can be cleaned up to be only Polar-related code, which will be much easier to review.  While I am in here, I also haul in PyFTDI.

Since absolutely fucking nothing can be easy, this system does not have `ldconfig`, which `ctypes.util.find_library` uses, so you need a special dance to make libusb work in Python:

```python
# workaround for missing ldconfig
import os
def find_library(lib):
    p = f"/usr/lib/{lib}.so"
    if os.path.exists(p):
        return p
import usb.backend.libusb1
usb.backend.libusb1.get_backend(find_library=find_library)

# now do any PyFTDI stuff you want...
```

cc: @ima-hima


riptidewave93 (approved):

ima-hima:
"I split the Python bits into their own PR, and rebuild Pycryptodome to target arm32 rather than Darwin."

Aargh! Sorry!

PR #294, opened 2024-06-10
Make OTA Enabled by Default

scooterthedev:
The OTA engine was disabled by default, so I made it enabled.


jwise:
this also needs to be modified in the printer_ui settings switch, I believe?

PR #298, opened 2024-06-11
ota: default to enabled

jwise:
For some reason, originally we thought that the OTA engine should default to disabled.  This is clearly wrong.  It should default to enabled.  While we're in there, remove the other confusing defaults that are not useful and just add clutter and confusion.


jwise:
Wow, two fixes for the price of one!  Tested on my printer.

jphannifan (approved):

PR #308, opened 2024-06-16
x1plusd MQTT service

jphannifan:
1) I've added the packages `paho-mqtt` and `aiomqtt` to the site-packages folder. `paho-mqtt` can be used to communicate with other wireless devices via MQTT or to listen and publish to local MQTT. `aiomqtt` is an asyncio wrapper for paho, which allows us to use these libraries directly with x1plusd. See [here](https://sbtinstruments.github.io/aiomqtt/) for documentation. I have added an example script with an extremely simple example for development outside of x1plusd. 

2) Implemented the backend for a x1plusd service, MQTTService. Currently there are 3 CLI functions:
check status: `x1plus mqtt status`
publish: `x1plus mqtt publish '{"print":{"sequence_id":0,"command":"gcode_line","param":"M106 P1 S255"}}'`
subscribe (wildcards allowed): `x1plus mqtt subscribe device/#`

Moving forward we need to figure out the following:
- How should custom topics be configured?
- Message filtering for subscriptions ([aiomqtt](https://sbtinstruments.github.io/aiomqtt/subscribing-to-a-topic.html#filtering-messages) has filtering features that may be of use)


images/cfw/opt/x1plus/lib/python/x1plus/client/mqtt.py
```diff
@@ -0,0 +1,184 @@
...
+
+        try:
+            payload = func(**kwargs)
+            print(payload)
+            topic = f"device/{sn}/request"
+            result = publish_message(topic,payload)
+            if result:
+                print(f"Successfully published to topic {topic}. Payload: {json.loads(payload)}")
+            else:
+                print("Failed to publish message")
```
jwise, line 115: I mention some in my broader review comments, but I don't think this is the right approach.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/handler.py
```diff
@@ -0,0 +1,66 @@
...
+    # this is pure dogshit and I am going to delete it as soon as it's clear how we want to set this up
+    async def x1plus_update(self, message):
+        topic = message.topic.value
+        logger.info(f"Received update message on {topic}")
+        try:
+            cmd = message.payload.decode('utf-8').strip().split(" ")
+            resp = self.subproc(["x1plus", "ota", cmd[0]] + cmd[1:])
+            await self.publisher("x1plus/update", resp)
+        except Exception as e:
+            logger.error(f"Error handling update command: {e}")
```
jwise, line 32: This I really like, and is something that I kind of hadn't considered.  It sort of seems to me like there are actually three MQTT-related things that are probably separate arcs of work:

* Getting data that normally is broadcast to the MQTT (or DDS) bus into x1plus client apps on the system;
* Having x1plus client apps on the system make requests over the system MQTT (or DDS) bus; and,
* Getting x1plus internal state (OTA info, settings changes, etc.) connected so that other MQTT users (like HA) can get access to them.

I'm wondering if, as gross as it is, we should continue to use the standard Bambu MQTT approach -- broadcast our announcements on `device/{sn}/response` and take commands on `device/{sn}/request`.  If nothing else, MQTT users are used to that at this point.  We could publish them with our own key, `"x1plus_ota": { "command": "status" }`, or whatever, though.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/mqtt.py
```diff
@@ -0,0 +1,180 @@
...
+        self.x1psettings = settings
+        self.broker_url = '127.0.0.1'
+        self.port = 8883
+        self.topic = f"device/{sn}/request"
+        self.username = 'bblp'
+        self.password = access_code
+        self.cafile = '/usr/etc/system/certs/ssl-ca-bbl.pem'
+        self.client_config = self.build_client_config()
+        self.client = aiomqtt.Client(**self.client_config)
+        self.status = self.STATUS_DISCONNECTED
```
jwise, line 33: This is useful.  I am going to yoink a bunch of this for my attempt at this.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/mqtt.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            self.enable = True
+            config = self.build_client_config()
+            self.client = aiomqtt.Client(**config)
+            await self.client.__aenter__()
+            self.status = self.STATUS_CONNECTED
+            logger.info("Connected to MQTT Broker, starting subscription...")
+            await self.subscribe_topics()
+            asyncio.create_task(self.listen_for_messages())
+            asyncio.create_task(self._outgoing())
```
jwise, line 83: why does this need to be a task?

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/mqtt.py
```diff
@@ -0,0 +1,180 @@
...
+            await self.publish(topic, json.dumps(message).encode('utf-8'))
+            return {"status": "ok"}
+        except Exception as e:
+            logger.error(f"Failed to publish message: {str(e)}")
+            return {"status": "error", "message": str(e)}
+
+    async def dbus_MQTTSubscribe(self, req):
+        topic = req.get('topic')
+        self.register_topic_handler(topic, self.default_handler)
+        return {"status": "ok"}
```
jwise, line 163: This definitely is an API that does not make a whole lot of sense :)

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
```diff
@@ -7,12 +7,22 @@
...
     Used to get the Serial Number for the Printer
     """
-    return subprocess.check_output(["bbl_3dpsn"], stderr=subprocess.DEVNULL).decode(
-        "utf-8"
-    )
+    if os.path.isfile("/oem/device/sn"):
+        return open("/oem/device/sn", "r").read()
+    else:
+        if os.system("bbl_3dpsn 2>/dev/null > /tmp/.bambu_sn") != 0:
+            return open("/tmp/.bambu_sn", "r").read()
```
jwise, line 19: Wait why are we doing this bonkers `/tmp/.bambu_sn` thing again?  That is the whole point of the `subprocess.check_output(["bbl_3dpsn"])`, so we odn't have to create any other files.  And every supported X1Plus installation should be able to run `bbl_3dpsn`, there should be no need for `/oem/device/sn` at this point.

images/site-packages/aiomqtt/x1plus_example.py
```diff
@@ -0,0 +1,73 @@
...
+        ) as client:
+            await client.subscribe("device/#")
+            
+            async for message in client.messages:
+                topic = message.topic.value
+                msg = message.payload.decode("utf-8")
+                print(f"{topic}\n{msg}")
+            
+    except Exception as e:
+        print(f"Error subscribing: {e}")
```
jwise, line 53: this definitely should not live in site-packages!  site_packages is things we have vendored in and should only EXACTLY be external code!

jwise (commented):
So I took a look at this, and I think I have a few general architectural comments.

* I'm not sure that MQTT is a *raw* interface that needs to be exported over DBus.  If a Python client on the system wants to talk to MQTT, it should just do it using paho -- why does x1plusd need to do it?
* More broadly, something I am thinking about is that I want to export stable interfaces from x1plusd.  So, for instance, making an `x1plus.client` formulate MQTT requests of its own seems like an interface that does not make a whole lot of sense (it sort of pushes the problem further down).  What I'd really like is to have a rich `x1plus.client` API.  For instance, instead of having to do `x1plus.client.mqtt.publish_message('device/{sn}/request', { 'print': { 'sequence_id': '0', 'command': 'gcode_file', 'param': file_path } })`, I would much rather have an API that end users can call of the form `x1plus.client.print.gcode_file(file_path)` that would hide all the complexity of having to send a command, and subscribe for the response, and get the response back to see if it succeeded.
* The question then is: how should that be implemented?  I guess if `x1plus.client` goes and creates an MQTT client of itself to do this, that's an implementation detail that I guess I'd be fine with.  But another option is, if the `x1plusd` is going to have a MQTT connection already, maybe there should be a DBus API `x1plus.print.PrintGcode({ filename = whatever })` that did that operation on your behalf.
* If you do that, you have a lovely bonus.  If the `x1plusd` has a long-lived MQTT connection that it's subscribed to all the time, then you can ask the `x1plusd` questions about print status without having to 'wait around' for the answers.  For instance, if `x1plusd` remembers the last printer status blob that it got over MQTT, you could quickly call up `x1plus.print.GetStatus({})` and get back everything that `x1plusd` knows.

So I have a kind of different idea for the X1Plusd mqtt service, and that is to have it export a 'listener' mechanism that you can use anywhere inside x1plusd.  If you want to send a request and listen for a response a short period of time later, you can do that -- and if you want to listen for all MQTT messages (for instance, because you are an httpd and you want to forward that to a websocket), you can do that too.  I'm going to try prototyping that.

I have a couple of other architectural comments that I left inline, as we think about how to define a MQTT outbound interface for X1Plus, also.

images/site-packages/aiomqtt/x1plus_example.py
jphannifan: yea this definitely was not supposed to be there! I do have an actual example that's useful, but it doesn't belong here

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
jphannifan: Oh ok, I was just copy pasting what I saw used in another one of our Python scripts (although I have now forgotten which one).

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/mqtt.py
jphannifan: It's one of those features that turned into a convenient debugging option so I haven't removed it yet. I don't think there's any value to users in subscribing to topics over command line. 

I do think it would be useful to have a simple interface for creating subscriptions, so that users are able to set up MQTT communication with esp32 or other devices, but this ain't it

images/cfw/opt/x1plus/lib/python/x1plus/client/mqtt.py
jphannifan: I've been trying to figure out what to do with this, and it's kinda morphed into something that's out of control. 

What I'm thinking is this whole option (_cmd_publish()) needs to be axed and replaced with CLI options that are more focused to the application. For example for gcode publication and file printing
```
x1plus print gcode_line "M420 S69"
x1plus print gcode_file /sdcard/test.gcode
```

These other things definitely should not be command line options, for multiple reasons

jphannifan:
Thanks for the feedback!

> * More broadly, something I am thinking about is that I want to export stable interfaces from x1plusd.  So, for instance, making an `x1plus.client` formulate MQTT requests of its own seems like an interface that does not make a whole lot of sense (it sort of pushes the problem further down).  What I'd really like is to have a rich `x1plus.client` API.  For instance, instead of having to do `x1plus.client.mqtt.publish_message('device/{sn}/request', { 'print': { 'sequence_id': '0', 'command': 'gcode_file', 'param': file_path } })`, I would much rather have an API that end users can call of the form `x1plus.client.print.gcode_file(file_path)` that would hide all the complexity of having to send a command, and subscribe for the response, and get the response back to see if it succeeded.

I fully agree. A lot of these things turned into convenient debugging features but I don't think they have much purpose beyond that. I guess the way I was looking at it is the CLI is easy to change around, so that exact implementation is not as important as is establishing a functional Python API. And that is something I am still pondering on. 

> * If you do that, you have a lovely bonus.  If the `x1plusd` has a long-lived MQTT connection that it's subscribed to all the time, then you can ask the `x1plusd` questions about print status without having to 'wait around' for the answers.  For instance, if `x1plusd` remembers the last printer status blob that it got over MQTT, you could quickly call up `x1plus.print.GetStatus({})` and get back everything that `x1plusd` knows.

We have quite a bit of freedom to choose here, because having an MQTT client always running in the background consumes very little memory (assuming it's rate limited/throttled which I have not done here). Since it also doesn't interfere with Bambu's client, from what I've seen so far, I see no real downsides to running a client that's always alive. 

> So I have a kind of different idea for the X1Plusd mqtt service, and that is to have it export a 'listener' mechanism that you can use anywhere inside x1plusd. If you want to send a request and listen for a response a short period of time later, you can do that -- and if you want to listen for all MQTT messages (for instance, because you are an httpd and you want to forward that to a websocket), you can do that too. I'm going to try prototyping that.

I think websocket integration will be one of the best features to come out of this PR. These next few updates will be huge

PR #310, opened 2024-06-17
Only allow installation on modern X1 base firmware.

jwise:
Over the last few weeks, there have been a handful of incidents that took up a decent amount of X1Plus support bandwidth associated with truly ancient base firmware making trouble for the X1Plus install process or runtime (in one case, a user ran across the known Bambu Lab RPMB migration bug associated with older firmware -- or had the wrong trustlets running -- and, in the other case, the init sequence in an ancient version of X1 firmware was different enough to break the WiFi setup process in the X1Plus installer).  We solve this by enforcing at install time that X1Plus must be installed on top of either 01.07.02.00 or the Official Rootable Firmware (which is 01.07.02.00 "plus a little extra stuff").  As an added bonus, this now will have an error at the beginning, rather than the end, of installation, if the on-eMMC firmware does not match a known version.  This will impact very few users (only users who previously installed using the exploit, on an ancient version); users who are affected will be given a clear failure message with a path forward.  (Technically savvy users who do not wish to upgrade to the Official Rootable Firmware may compile their own versions of X1Plus that do not have this warning; this configuration would be clearly unsupported.)

### DIFF
--- a/installer/install.py
+++ b/installer/install.py
@@ -47,7 +47,7 @@ basefw_mtime = INFO["base"]["mtime"]
 cfw_version = INFO["cfwVersion"]
 cfw_squashfs = f"{cfw_version}.squashfs"
 demand_free_space = fs_size + 1 * 1024 * 1024 * 1024
-# latest_safe_bblap = "00.00.28.55" # 1.07.02.00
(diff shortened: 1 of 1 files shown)


riptidewave93 (approved):

bani6809 (approved):
tested, seems to work

PR #312, opened 2024-06-19
WiFi status indicator for kexec_ui

jphannifan:
This uses a QProcess and shell script to check `/config/device/conn_mode` and `/proc/net/wireless` to see if the printer is in LAN mode or if it's active and reports a signal strength. The type is initialized in Screen.qml so it appears everywhere in kexec_ui. Updates are handled with bindings.

### DIFF
--- a/bbl_screen-patch/kexec_ui/printerui/qml/Screen.qml
+++ b/bbl_screen-patch/kexec_ui/printerui/qml/Screen.qml
@@ -33,7 +33,30 @@ Rectangle {
     height: 720
     color: Colors.gray_500
     
+    WifiStatus {
+        id: _wifiStatus
+    }  
+    
+    Text {
+        id: wifiLevelTxt
+        z: 1
+        anchors.top: parent.top
+        anchors.right: parent.right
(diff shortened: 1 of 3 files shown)


jwise:
I have not tried this on my printer, but, roughly:
* I do not understand why the *boot screen* should have the clutter of the WiFi signal strength, in dBm?  Most of the time, the printer will never use WiFi when in kexec_ui.
* We should not dynamically create shell scripts from QML.  They need to be packed in, if they are static things.
* LAN mode does not matter in kexec_ui.  If the WiFi is on, and we have permission from the user, we will use it.  If it is not, we will not.  There is no need to confuse the user by saying that we are in "LAN mode" while we could very well still be accessing the network.
* I am not sure why this needs to exist, in general.  If there is sufficient signal strength, then we will download -- otherwise, we will error.  And, anyway, now the OTA engine means that we almost never download from kexec_ui.

jphannifan:
This PR aged like milk after OTA updates went live. Who do I contact to file for a time refund?

PR #313, opened 2024-06-19
use netService in installer instead of wpa_supplicant_hook

### LINKED ISSUE #309: Upgrade to 2.0_RC0 fails on network unreacheable
**Describe the bug**
During install, installer locks up trying to download firmware.
Wifi connection is lost.
After that, sd card is no longer bootable.

**To Reproduce**
1. power down printer
2. remove SD card
3. put x1plus 2.0_RC0 x1p file on SDcard root
4. insert SD card
5. power on printer
6. chose 'Start x1plus install'
7. chose 2.0_RC0
8. wait for installer to unpack and start install process
9. printer reports wifi disconnect and gets stuck on downloading firmware.
10. reboot or powercyle printer
11. go back to step 1

**Expected behavior**
Full install of X1C 2.0_RC0

**Screenshots/Photos**
If applicable, add screenshots or photos to help explain your problem.

**External Links**
Has this bug been discussed elsewhere that you can link to? Have you checked this repository's discussions section?

**Additional Details:**
 - Printer type (X1/X1C/X1E): X1C
 - X1Plus version: 1.1
 - AP firmware version: 
 - AP hardware revision:
 - MC firmware version:
 - MC hardware revision:
 - Toolhead firmware version:
 - Toolhead hardware revision:
 - Computer vision model:

If it is, which version? *ex. 01.07.02.00*
Are any steps to replicate it different?

**Custom modifications**
Add details of any software or electronic modifications you've made to your printer here.

**Additional context**

jwise:
There have been some reports that WiFi works on OEM and works in X1Plus, but not in the installer -- e.g., #309.  In modern OEM firmware (and, indeed, in X1Plus!), the network is managed by `/usr/bin/netService`, rather than by `wpa_supplicant_hook`.  And, in fact, `wpa_supplicant_hook` sometimes can use the wrong config file (though, oddly, in Firmware R, `netService` does not use the "wpa_supplicant_new.conf" like it does in the X1Plus base firmware, but all the same, the user in #309 reports that WiFi works booted into OEM, so ...).

We resolve this in two ways:
* From now on, for all new `kexec_ui`s, we start `netService` instead of `wpa_supplicant_hook` for the early-start WiFi initialization.
* For `install.py` instances running on old `kexec_ui`s, if we detect that `wpa_supplicant_hook` is running, we kill it and restart into `netService`.

While we're in there, we add a little bit of version logging around `install.py`; there are a lot of builds floating around to try to debug this issue, and I realized that we have no way of knowing what build a given install log came from.

Fixes #309.  Hopefully.


bani6809 (approved):
tested, seems to work

PR #316, opened 2024-06-20
Bump translations

jwise:
Some things had changed in how CrowdIn generates their zip file, so that needed to be fixed.  After that, bumped translations forward.  Some late-breaking strings might not end up translated in 2.0.


riptidewave93 (approved):

jphannifan:
When I encountered this, I couldn't find any old Crowdin builds that I could download to see how the filenames had changed so I left it in a broken state. Thanks for taking care of this

PR #318, opened 2024-06-21
Polar/add settings

ima-hima:
Added use of Settings module into Polar Cloud. Also added a rudimentary status function that sends the status of the printer three times a minute to the cloud. A little code cleanup. Fixed the crypto signing algorithm.


ima-hima:
I'm killing this pull request, as it was against dev/v2.1, which has now been merged into main. Instead polar/add-dbus will take its place.

PR #320, opened 2024-06-23
Add support for USB Ethernet adapters on X1 Carbon

jwise:
In theory, only the X1E has support for Ethernet.  As it turns out, vestiges of this support live in the X1 Carbon firmware, but it needs some help.  To help it out, we:

* Interpose `libbbl_utils`, which, on X1 Carbon, doesn't know that `eth0` being up means that the network could be up.  `device_gate` needs to know this in order to be willing to connect to the cloud over Ethernet (thanks @balika011).
* Binary-patch `netService`, which doesn't know to create an `eth0` route on X1 Carbon, but can be convinced to with a little bit of help (thanks @balika011).
* Look for the X1Plus Expansion Board on boot, and if one is connected that doesn't have a MAC address in its EEPROM, assign an LAA based on the printer serial number so we get a consistent IP address each time we boot.
* Write some QML to listen to `netService`, because `bbl_screen` does not know how to listen for a `NetworkManager.wiredNetwork`; we create an `X1Plus.Network.wiredNetwork()` interface that looks a lot like a `NetworkManager.wiredNetwork` in response to `netService` DDS events, and replace JavaScript lookups of `NetworkManager.wiredNetwork` to point to this interface.

This allows X1 Carbons augmented with a USB Ethernet adapter of any kind (including an X1Plus Expansion Board) to use the native X1E Ethernet infrastructure to connect instead of WiFi, and provides the X1E GUI when an `eth0` interface of some kind is present.

### DIFF
--- a/bbl_screen-patch/Makefile
+++ b/bbl_screen-patch/Makefile
@@ -7,7 +7,7 @@ TARGET_CC ?= $(TOOLCHAIN)gcc
 LDFLAGS = -shared -ldl
 CXXFLAGS = -fPIC -std=c++11 -Ivendor -O3 -mcpu=cortex-a7 # yes, -O3 actually helps here with the VNC shenanigans
 CFLAGS = -fPIC -Ivendor -O3 -mcpu=cortex-a7 # yes, -O3 actually helps here with the VNC shenanigans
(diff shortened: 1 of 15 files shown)


bani6809 (approved):
it seem to work with usb ethernet dongle

PR #329, opened 2024-06-27
Merge dev/v2.1 onto main

jwise:
2.0 is out the door, so we can develop on main now.

This will be merged with a rebase, not a squash!


bani6809 (approved):
seems to print and not asplode. approved

PR #332, opened 2024-06-28
Polar/add dbus

ima-hima:
Rebases on main.

This adds `SettingsService` and `DbusService` into `PolarPrintService`.


ima-hima:
Closing this because every push is spamming Discord.

PR #340, opened 2024-07-04
Vendor in aiomqtt-2.2.0 and paho_mqtt-2.1.0.

jwise:


jphannifan (approved):
Oh nice, I see there's been a recent release for aiomqtt. 

On an unrelated note, are you familiar with AnyIO? I was looking at the aiomqtt repo and realized that they support AnyIO. There are some networking features that imo improve how sockets are handled in asyncio Python. Maybe we will find a use for it: https://github.com/agronholm/anyio

PR #343, opened 2024-07-08
Print Speed UI

jphannifan:
This branch was a month behind main, and there are only two files so it's easier to cherry pick. 

Summary: SpeedAdjust.qml allows you to adjust your print speed from 30 to 180% at a step size of 2%. This is a Dial type that is annotated by a canvas plot. It started out as a slider, and it was hard to control so now it's a dial. This currently functions as an additional speed adjust UI to the OEM stepbar. 

This is a screenshot of the most recent update to the UI. Switching between the OEM and X1Plus speed UIs is managed by X1Plus setting "printerui.speedadjust" (bool). The default value for this setting ensures that the OEM speed UI is always displayed the first time. I guess we can call it an 'opt-in' experimental feature

interpolate

Here are some useful resources associated with this branch:
- [Gcode speed function](https://github.com/X1Plus/X1Plus/blob/d7038787ac6f1d69373a80e792b69d40d5673874/bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js#L521) - PR in which I published speed adjust Gcode function 
- [My documentation of interpolation methods](https://github.com/jphannifan/x1plus-testing/blob/main/BL-speed-adjust.md)

Also @jwise you may have already seen this but I made several revisions after you reviewed the previous branch, and I believe it clears up some of your concerns. The only outstanding issue I really see here is the small discrepancy between Bambu parameters and calculated parameters at 100%. See the plot below for a detailed look:
interpolate


riptidewave93 (approved):

bbl_screen-patch/patches/printerui/qml/printer/SpeedAdjust.qml
```diff
@@ -0,0 +1,385 @@
...
+            anchors.top: parent.top
+            anchors.topMargin: 40
+        }
+        Dial {
+            id: dial
+            width: 300
+            height: width
+            from: 30
+            stepSize:2
+            to:180
```
jwise, line 140: This is 50 to 166 now, since we don't extrapolate outside of the Bambu domain, right?

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
```diff
@@ -490,62 +490,96 @@ const Tramming = {
...
-            feedRate: feedRate,
-            level: level > 7 ? 7 : level
+    }
+
+    function scaleFn(func, dataPoints) {
+        return (x) => {
+            const speeds = dataPoints.map(point => point[0]);
+            const values = dataPoints.map(point => point[1]);
+            const scalingFactor = interpolate(x, speeds, values) / func(x);
+            return func(x) * scalingFactor;
```
jwise, line 543: haha wait dividing by func(x) and then multiplying through by func(x) is just the same as multiplying by 1, right?  so the func(x) is not used here at all, and this *only* does the linear interpolation, right?

bbl_screen-patch/patches/printerui/qml/printer/SpeedAdjust.qml
```diff
@@ -0,0 +1,385 @@
...
+            
+        }
+        
+        Text { 
+            id: txtParams //displays the parameters that will be applied
+            property var speedParams: gcode.printSpeed(dial.value)
+            text:qsTr("Acceleration:\n%1\nFeed rate:\n%2\nTime:\n%3")
+                .arg(speedParams.accelerationMagnitude.toFixed(2))
+                .arg(speedParams.feedRate.toFixed(2))
+                .arg(speedParams.speedFraction.toFixed(1))
```
jwise, line 297: Maybe the following might be more understandable?
```suggestion
            text:qsTr("Acceleration:\n%1%%\nFeed rate:\n%2%%\nTime:\n%3%%")
                .arg((speedParams.accelerationMagnitude*100.0).toFixed(0))
                .arg((speedParams.feedRate*100.0).toFixed(0))
                .arg((speedParams.speedFraction*100.0).toFixed(0)) 
```

bbl_screen-patch/patches/printerui/qml/printer/AdjustPage.qml.patch
```diff
@@ -1,17 +1,29 @@
...
+@@ -16,7 +17,13 @@
      property var extruders: hasExtruderList ? extruderList.extruders : []
      property var currentExtruder: hasExtruderList ? extruderList.currentExtruder : null
      property var tmpModelArray: hasExtruderList ? (extruders.map(n=>n.extruderHeater)).concat(tempModel) : [heaters.hotend].concat(tempModel)
 -
 +    property var isHomed: (PrintManager.homedState & PrintManager.AXIS_HOMED_ALL) >= PrintManager.AXIS_HOMED_ALL
++    
 +    property bool ignoreDialog: false
++    property var speedValues: [50,100,124,166] // corresponding to PrintManager.currentTask.printSpeedModeStrings
++    property var speedStr: [3,2,1,0].map(n => PrintManager.currentTask.printSpeedModeStrings[n])
```
jwise, line 20: I think this is unused now?
```suggestion
```

bbl_screen-patch/patches/printerui/qml/printer/SpeedAdjust.qml
```diff
@@ -0,0 +1,385 @@
...
+
+Item {
+    id: speedDial
+    width: parent.width
+    height: parent.height
+    anchors.fill: parent
+    property alias speed: dial.value
+    property var targetSpeed:100
+    property var currentSpeed: PrintManager.currentTask.printSpeed
+    property var speedLabel:  (Math.abs(currentSpeed-targetSpeed) > 0) ? `${targetSpeed} / ${currentSpeed}%` : `${currentSpeed}%`
```
jwise, line 38: 1) I believe that conditional is usually spelled `currentSpeed != targetSpeed`.
2) I'm sort of thinking that maybe we do not need that anyway?  Like maybe this should always be `${currentSpeed}%`, since it reflects what state the printer is in right now, and the change only takes place when you click "apply"?
```suggestion
    property var speedLabel: `${currentSpeed}%`
```

jwise (approved):
Looks pretty good!  I think I would be good with this, though I would appreciate if you would fix the few nits that I highlighted.  The other bug I noticed is that if you click on the speed adjust and then click away from it, then it dehighlights:

image

But if you click on it, and then click apply, then it stays highlighted: 

image

I think the `tempList.currentIndex` isn't getting reset to `-1` in that case?

Also I seriously do not understand how `PrintManager.currentTask.printSpeed` ends up getting set to the correct percent.  Like when I set it to 136%, the Gcode published is M73.2 R0.73, which is OK, because 1/0.73 is 1.369, and the floor of that I guess is 136%.  But when I set it to 132%, the Gcode published is M73.2 R0.75, and 1/0.75 rounds to 133.3%!  Does it, like, floor() to the nearest 2% or something?  How does it... know?

Thanks for pushing this all the way through.  It's cool to see people really wanting this in the field.  Let's get it to them!

jphannifan:
> Looks pretty good! I think I would be good with this, though I would appreciate if you would fix the few nits that I highlighted. The other bug I noticed is that if you click on the speed adjust and then click away from it, then it dehighlights:
> 

Thanks for taking a look and I appreciate the feedback. I have a couple improvements I want to make and I have revisions for each of your comments, and I'll wrap that up today. 

> I think the `tempList.currentIndex` isn't getting reset to `-1` in that case?

I see the issue here. The 'Apply' button dismisses speedPad by setting speedPad.target to null, which prevents it from resetting the property you mentioned. 

> Also I seriously do not understand how `PrintManager.currentTask.printSpeed` ends up getting set to the correct percent. Like when I set it to 136%, the Gcode published is M73.2 R0.73, which is OK, because 1/0.73 is 1.369, and the floor of that I guess is 136%. But when I set it to 132%, the Gcode published is M73.2 R0.75, and 1/0.75 rounds to 133.3%! Does it, like, floor() to the nearest 2% or something? How does it... know?

This one makes no sense to me either.. First off it does the calculation on the MC board apparently, and this value is shared with the slicer via MQTT. I guess since it's calculating the time remaining, that is most accurately done on the MC board. I haven't found a single value that will result in an odd-numbered print speed. It's always rounded down to the closest even. Second this suggests that the printer is very confused about what time it is. You can adjust `M73.2 R` to make it report anywhere from 0% to 500% regardless of the other parameters. I'm glad you convinced me to correct the calculations especially after realizing this. 

> 
> Thanks for pushing this all the way through. It's cool to see people really wanting this in the field. Let's get it to them!

Took me long than it should have but I think we could have a release candidate ready within the next month or two maybe? Depending on what the timeline is for the expansion board

bbl_screen-patch/patches/printerui/qml/printer/SpeedAdjust.qml
jphannifan: The main reason this exists is because I had initially made a "ramp" feature that would adjust speed at layer changes. Since there is no case now where `targetSpeed` should differ from `currentSpeed`, it's definitely not necessary.

For now I'm saving the ramp feature for later. I think we nearly have a framework that makes `onLayerChange` gcode macros fairly easy to do

bbl_screen-patch/patches/printerui/qml/printer/SpeedAdjust.qml
jphannifan: I think I'm also going to add a tooltip button in the upper right corner so that users don't have to hunt for documentation to figure out what these numbers mean

bbl_screen-patch/patches/printerui/qml/x1plus/GcodeGenerator.js
jphannifan: The ratio is between linear interpolation of the "theoretical" value vs the linear interpolation of the "actual" value. 

This happened by accident, and I am now realizing I should have just used cubic spline interpolation because interpolating interpolated data is not exactly logical to me. 

This results in a smoothed linear interpolation very similar to cubic spline interpolation 
image

Glod76:
is this going to be added to main soon?

jb1228:
@jwise Is there anything else that needs to be done with this PR?

For whatever reason (hopefully nothing too bad 😟), there has been no activity from @jphannifan since 2024.

PR #345, opened 2024-07-13
x1plusd: add MQTT client

jwise:
This PR adds an MQTT client using `aiohttp` to `x1plusd`, and an internal interface to subscribe to MQTT messages.  It adds a sample client that subscribes to MQTT messages and forwards them to the httpd websocket.  The MQTT client can be configured to connect to a non-local MQTT server (for instance, if you are running `x1plusd` in emulation, and want it to talk to a real printer for status) using the `mqtt.override.*` settings.

For instance, when running in emulation, consider:

```
$ PYTHONPATH=../lib/python/ ./x1plus settings get
http.bind.port: 8080
http.enabled: true
http.password: "lol"
mqtt.override.host: "10.1.10.50"
mqtt.override.password: "[printer access code]"
mqtt.override.sn: "[printer serial number]"
$ websocat ws://127.0.0.1:8080/ws
{"jsonrpc": "2.0", "method": "hello", "params": {"x1plus_protocol_version": 0}}
{"jsonrpc": "2.0", "method": "auth", "params": {"password": "lol"}, "id": 0}
{"jsonrpc": "2.0", "result": 0, "id": 0}
```

... and you should begin receiving `mqtt.notify` jsonrpc events.  

There are a handful of useful things to do from here as future work:
* `x1plusd` probably should gain an internal status accumulator that listens to messages over MQTT, and buffers the answers as needed to be able to answer over DBus to an `x1plus.client` that wants to know how things are going.
* The Polar Cloud plugin should use the buffered status accumulator to report status to the cloud, and should subscribe to messages.
* The "wait-multiple-queues-or-generators" thing probably is going to be useful in a way that should be refactored.
* More proximately, I wanted to use the status indication in order to drive LEDs on the expansion board.  So I'm going to do that next.

Thanks to @jphannifan for early work on figuring out how to drive `aiomqtt`!

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
@@ -8,6 +8,7 @@ from .settings import SettingsService
 from .ota import OTAService
 from .sshd import SSHService
 from .httpd import HTTPService
+from .mqtt import MQTTClient
 
 logger = logging.getLogger(__name__)
 
@@ -17,10 +18,12 @@ async def main():
 
     router = await get_dbus_router()
     settings = SettingsService(router=router)
+    mqtt = MQTTClient(settings=settings)
     ota = OTAService(router=router, settings=settings)
     ssh = SSHService(settings=settings)
     httpd = HTTPService(router=router, settings=settings)
 
+    asyncio.create_task(mqtt.task())
     asyncio.create_task(settings.task())
     asyncio.create_task(ota.task())
     asyncio.create_task(httpd.task())
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
@@ -21,7 +21,7 @@ async def main():
     mqtt = MQTTClient(settings=settings)
     ota = OTAService(router=router, settings=settings)
     ssh = SSHService(settings=settings)
-    httpd = HTTPService(router=router, settings=settings)
+    httpd = HTTPService(router=router, settings=settings, mqtt=mqtt)
 
     asyncio.create_task(mqtt.task())
     asyncio.create_task(settings.task())
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/httpd.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/httpd.py
@@ -16,13 +16,14 @@ access_logger = logging.getLogger(__name__ + ".access")
 PROTOCOL_VERSION = 0
 
 class HTTPService():
-    def __init__(self, settings, router, **kwargs):
+    def __init__(self, settings, router, mqtt, **kwargs):
         self.x1psettings = settings
         self.x1psettings.on("http.enabled", lambda: asyncio.create_task(self.sync_startstop()))
         self.router = router
         self.runner = None
         self.site = None
         self.app = web.Application()
+        self.mqtt = mqtt
         
         if x1plus.utils.is_emulating():
             static_path = Path(__file__).resolve().parent.parent.parent.parent.parent.parent / 'share' / 'www'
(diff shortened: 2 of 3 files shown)


riptidewave93 (approved):

jphannifan (approved):
+1 nice work

I like the idea of having override options for client setup, but if the goal is to allow communication with other devices, isn't there a need to allow topic overrides? Say I want to communicate with an esp32, unless I've configured the MQTT client on the esp32 myself so it publishes to device/+/request then I'm out of luck (well in reality I'd just write some python and it wouldn't be a crisis)

A couple other small nitpicks - 

1) Handling disconnections - I think this needs a way to ensure that outbound messages aren't lost during temporary disconnections. I used an outbound message queue in my branch and this seemed to serve its purpose 

2) I think it might be a good idea to use the QoS argument when publishing MQTT messages, just to ensure that QoS=1 is always used.

jwise:
@jphannifan wrote:

> I like the idea of having override options for client setup, but if the goal is to allow communication with other devices, isn't there a need to allow topic overrides? Say I want to communicate with an esp32, unless I've configured the MQTT client on the esp32 myself so it publishes to device/+/request then I'm out of luck (well in reality I'd just write some python and it wouldn't be a crisis)

So this is *only* supposed to be used to connect to the printer's inbuilt MQTT broker.  Connecting to other MQTT brokers is out of the scope of this.  If you want to connect to another device, write a different program (ideally on that device, but if need be, as an x1plusd client would be fine too).

>     1. Handling disconnections - I think this needs a way to ensure that outbound messages aren't lost during temporary disconnections. I used an outbound message queue in my branch and this seemed to serve its purpose

Temporary disconnections shouldn't happen -- if the printer's built in MQTT broker crashes, then something else has gone wrong with the printer.  (Non-local connections are only ever used in development; this client will only ever connect to the printer over loopback in a real printer.)

I'm not sure if the right thing to do is to queue or to drop the message if there is a broker fault.  Queuing can be bad because it means that a message can be delivered (much?) later if the broker finally wakes up, perhaps when it is no longer relevant (see, for instance, the queuing problems that Bambu have had during their cloud outages, where printers wake up and start printing in the middle of the night!).

By contrast, queuing also may not *solve* problems.  If we redeliver just as the MQTT broker starts up, the other MQTT client with which we are attempting to communicate may not have connected yet, so we might send a message, but there is nobody to receive it!

It is best to consider MQTT as being a best-effort protocol -- this is why Bambu uses sequence numbers and responses (though, to my knowledge, they do not deduplicate retransmitted sequence numbers!).

>     2. I think it might be a good idea to use the QoS argument when publishing MQTT messages, just to ensure that QoS=1 is always used.

QoS=1 only works, by my reading of the spec, for subscribers that are already connected, right?  So for instance if we end up in the same condition as above, where the mqtt broker crashes and we are the first to reconnect, then we can send a QoS=1 message, but it still might not be delivered to whomever we actually care about delivering it to.  For this, the same behavior as the above problem is how this is solved:

* for messages that x1plusd publishes to the `request` endpoint, to make the printer do something in particular, the caller should wait until it sees the response on the `report` endpoint before declaring victory
* messages published to the `report` endpoint are, by their nature, best-effort; if they are publishing some state to the outside world, either the outside world will rerequest the action if it wanted something to happen, or if it is periodically publishing state, the outside world will be able to wait for the next publish interval.

Does this make more sense?

jphannifan:
> So this is _only_ supposed to be used to connect to the printer's inbuilt MQTT broker. Connecting to other MQTT brokers is out of the scope of this. If you want to connect to another device, write a different program (ideally on that device, but if need be, as an x1plusd client would be fine too).

What are the overrides for then? I agree that this should only be handling local (on printer) MQTT, because supporting other brokers would get pretty messy. So aside from debugging, what is the use case for the override settings? 

> Temporary disconnections shouldn't happen -- if the printer's built in MQTT broker crashes, then something else has gone wrong with the printer. (Non-local connections are only ever used in development; this client will only ever connect to the printer over loopback in a real printer.)

Fair enough. The main scenario in which I see this occurring is x1plusd crashing or restarting. It should not happen but it definitely will, and it sounds like maybe the error handling I have in mind can be done by the services/scripts that are using the client, rather than the client itself.

> QoS=1 only works, by my reading of the spec, for subscribers that are already connected, right? So for instance if we end up in the same condition as above, where the mqtt broker crashes and we are the first to reconnect, then we can send a QoS=1 message, but it still might not be delivered to whomever we actually care about delivering it to. For this, the same behavior as the above problem is how this is solved:

I had a senior moment there. I meant QoS=0. I know that Paho is _supposed_ to use a QoS of 0 by default, but I haven't done any checking into whether there are scenarios that this is not the case. 

> It is best to consider MQTT as being a best-effort protocol -- this is why Bambu uses sequence numbers and responses (though, to my knowledge, they do not deduplicate retransmitted sequence numbers!).

Do we not need sequence numbers or some form of message id to correctly match requests to responses? The MQTT client itself does not need sequence numbers to succeed, as long as QoS=0, but our client's message handlers may not function smoothly otherwise. It may be useful for debugging purposes to have some rudimentary form of message tracking

jwise:
> What are the overrides for then? I agree that this should only be handling local (on printer) MQTT, because supporting other brokers would get pretty messy. So aside from debugging, what is the use case for the override settings?

The overrides are so you can run x1plusd in emulation, and have it connect to a real printer's mqtt broker -- they are *only* for use in emulation / development.

> Fair enough. The main scenario in which I see this occurring is x1plusd crashing or restarting. It should not happen but it definitely will, but it sounds like maybe the error handling I have in mind can be done by the services/scripts that are using the client, rather than the client itself.

Yeah, indeed, this is services/scripts that should handle that.  If x1plusd crashes, then the queue gets purged by the time x1plusd comes back to life!

> I had a senior moment there. I meant QoS=0. I know that Paho is _supposed_ to use a QoS of 0 by default, but I haven't done any checking into whether there are scenarios that this is not the case.

I see that the `qos` kwarg defaults to 0 in the documentation, so I think it is safe to rely on this behavior:

image

> Do we not need sequence numbers or some form of message id to correctly match requests to responses? The MQTT client itself does not need sequence numbers to succeed, as long as QoS=0, but our client's message handlers may not function smoothly otherwise. It may be useful for debugging purposes to have some rudimentary form of message tracking

Indeed, if we introduce the ability to respond to something that someone publishes on the `request` topic, we should use message IDs the same way that Bambu does in our `report`s.

PR #348, opened 2024-07-19
fix translation error reported in https://discord.com/channels/1041781310299062282/1188616216936468542/1263997005240336426

jwise:


bani6809 (approved):
confirmed bug fixed

PR #350, opened 2024-07-29
Add X1Plus Expansion Board support to x1plusd.

jwise:
The [X1Plus Expansion Board](https://github.com/jwise/x1-expander) is a platform for X1 that allows external hardware features.  The initial revision, X1P-002-B01, has support for the following:

* An Ethernet port, provided by LAN9514.
* A small EEPROM with board identity.
* Up to four expansion ports (two ports on current builds with FT2232) for addin modules like X1P-004 and X1P-005.
* Two USB ports.
* An Adafruit / Sparkfun-compatible I2C connector.

We add support in x1plusd to detect the board's EEPROM, to detect addin module EEPROMs, and to support some I2C peripherals and an addin module.  To do that, we add the following:

* `x1plusd.expansion.manager` is added as scaffolding to respond to DBus requests to query the expansion state, and to probe the expansion board on boot.  Additionally, the expansion manager reacts to x1plusd.settings changes that specify what is attached to each port of the FTDI.
* We introduce drivers, `x1plusd.expansion.i2c` and `x1plusd.expansion.ledstrip`, that implement support for two different types of I2C thermometers (SHT41 and AHT20), and a few different animations on a WS2812B LED strip.
* We add an initial `x1plusd.sensors` manager -- for now, only holding values from I2C sensors, but in the future, it will also hold internal printer sensors.
* We make a small change to `x1plusd` to pass around a single `X1PlusDaemon` object, rather than each of the subobjects -- this makes it more possible to reach the MQTT subsystem from the sensors subsystem, or the sensors subsystem from the expansion subsystem.

To interface with this, consider settings of the following form:

```
[root@BL-P001-sdcard:~]# x1plus settings get
expansion.port_a: {"ledstrip": {"leds": 10}}
expansion.port_b: {"i2c": {"0x44": {"sht41": {"name": "room"}}}}
```

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
@@ -14,26 +14,36 @@ from .sensors import SensorsService
 
 logger = logging.getLogger(__name__)
 
-async def main():
-    logger.info("x1plusd is starting up")
+class X1PlusDaemon:
+    @classmethod
+    async def create(cls):
+        # must be a classmethod, since __init__ cannot be async
+        self = X1PlusDaemon()
 
-    router = await get_dbus_router()
-    settings = SettingsService(router=router)
-    mqtt = MQTTClient(settings=settings)
-    ota = OTAService(router=router, settings=settings)
-    ssh = SSHService(settings=settings)
-    httpd = HTTPService(router=router, settings=settings, mqtt=mqtt)
-    sensors = SensorsService(router=router, settings=settings)
-    expansion = ExpansionManager(settings=settings, sensors=sensors)
+        logger.info("creating x1plusd services")
+        self.router = await get_dbus_router()
+        self.settings = SettingsService(router=self.router)
+        self.mqtt = MQTTClient(daemon=self)
+        self.ota = OTAService(router=self.router, daemon=self)
+        self.ssh = SSHService(daemon=self)
+        self.httpd = HTTPService(router=self.router, daemon=self)
+        self.sensors = SensorsService(router=self.router, daemon=self)
+        self.expansion = ExpansionManager(daemon=self)
+        if not self.settings.get("polar_cloud", False):
+            self.polar_cloud = None
+        else:
+            from .polar_cloud import PolarPrintService
+            self.polar_cloud = PolarPrintService(daemon=self)
+        
+        return self
 
-    asyncio.create_task(mqtt.task())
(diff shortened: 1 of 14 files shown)


jphannifan:
This is truly incredible work! I look forward to getting my hands on a board so I can contribute.

I was looking through the sensors I have recently and found a few analog sensors like a [MICS6814](https://www.sgxsensortech.com/content/uploads/2015/02/1143_Datasheet-MiCS-6814-rev-8.pdf) gas sensor, many reels of [DRV5053](http://www.ti.com/lit/ds/symlink/drv5053.pdf) magnetometers, and an [ALSPT19](https://cdn-shop.adafruit.com/product-files/2748/2748%20datasheet.pdf) light sensor and many more than I can list here. I don't have the specific applications worked out for all of these, but it seems the first task here is finding a decent and inexpensive ADC. Ideally in breakout board form I think.

These [ADS1015](https://www.adafruit.com/product/1083) breakout boards look suitable to me. That's on the pricy side though. Have you thought about this at all? 

As for i2c devices, I think most users would be interested in accelerometers and humidity sensors. I own a few ADXL345 and MPU6050 accelerometers and they're very good cheap sensors. I like BME280 for humidity but there are a ton of good i2c hygrometers out there. Humidity sensing would be amazing for print farms

jphannifan (approved):

jwise:
ADS1015 sounds like a good option to me.  Adafruit stuff is always kind of pricey: the ADC itself is more like $1.  So if you wanted to do your own board with a bunch of sensors attached to an ADS1015, that is definitely workable.  But I think that people who want to add sensors to their X1 are probably not all that price-sensitive.  (I am counting on it, actually.)

I was also thinking about a light sensor.  I have one here that is directly connected by I2C.  I think the play there would be to automatically have the X1 turn its lights on and off when the room lights are on and off, for automatic bedtime.

jphannifan:
> So if you wanted to do your own board with a bunch of sensors attached to an ADS1015, that is definitely workable. 

I think for now I'm just going to find a cheap option for an ADC breakout and wait to see what the community is most interested in. I've seen the amount of time and energy you've put into developing the expansion boards, and I'm not sure how you do it. I think working on a few high quality projects for the board is the least we can do 

> I was also thinking about a light sensor. I have one here that is directly connected by I2C. I think the play there would be to automatically have the X1 turn its lights on and off when the room lights are on and off, for automatic bedtime.

This one with gesture sensing is kinda cool https://learn.adafruit.com/adafruit-apds9960-breakout

PR #357, opened 2024-08-22
x1plusd.expansion: avoid crash on startup with no expansion board installed

jwise:
oopsies.  thanks @ima-hima for the report


ima-hima (approved):
Tested and works.

PR #358, opened 2024-08-25
Updated README to include instructions for Apple Silicon.

ima-hima:
The Docker calls are different if you're on a Mac with new hardware. I added those instructions into the readme. This also pulls forward some Mac-specific Docker calls from an earlier PR that were approved but didn't get merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ about how to make changes and how to contribute back.  X1Plus is the result
 of a year or so of vaguely-structured work; it started life as a fairly
 frumious hack, and over the past few months, we've been working hard to try
 to clean it up for release to the outside world.  All that said, there are
-parts that you will find are still a mess, and for that, we're truly sorry! 
+parts that you will find are still a mess, and for that, we're truly sorry!
 We hope you'll come play around with it anyway.  There's a lot of fun stuff
 to be done, and we have only barely scratched the surface so far.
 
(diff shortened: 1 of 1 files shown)


jwise:
Hm OK I flipped through this and this is going to require a little thinking on my part for how to make this a little more coherent.

I think one side of it is that probably I want to separate the "how to set up the build system" from the mechanics of building X1Plus once you know how to drive Docker.  I think the other side of it, though, is that this description puts the `getkey` step before `getkey` is actually built by the Docker image :-) So I definitely need to figure out how to put that together, but that might not be a tonight project.

ima-hima:
> Hm OK I flipped through this and this is going to require a little thinking on my part for how to make this a little more coherent.
> 
> I think one side of it is that probably I want to separate the "how to set up the build system" from the mechanics of building X1Plus once you know how to drive Docker. I think the other side of it, though, is that this description puts the `getkey` step before `getkey` is actually built by the Docker image :-) So I definitely need to figure out how to put that together, but that might not be a tonight project.

You definitely have to have X1Plus installed before you can run `getKey`, right, because you need to be able to ssh in? Which is very Catch-22, but maybe using the installer to do that is just a pre-req for someone to start developing?

Tbh, I'm not sure happy about the organization of these, either. I think it's a little confusing the way I have the various instructions split up. And Windows is still missing, as well.

RaySajuuk:
This is what I use to build on Apple Silicon. Simpler?
```
docker build --platform linux/amd64 -t x1plusbuild scripts/docker/
docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild bash -c 'git config --global --add safe.directory /work'
docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild make scripts
docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild make
```

ima-hima:
> This is what I use to build on Apple Silicon. Simpler?
> 
> ```
> docker build --platform linux/amd64 -t x1plusbuild scripts/docker/
> docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild bash -c 'git config --global --add safe.directory /work'
> docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild make scripts
> docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild make
> ```

Hmm. Maybe, yeah. I'll do some testing.

RaySajuuk:
> > This is what I use to build on Apple Silicon. Simpler?
> > ```
> > docker build --platform linux/amd64 -t x1plusbuild scripts/docker/
> > docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild bash -c 'git config --global --add safe.directory /work'
> > docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild make scripts
> > docker run --platform linux/amd64 -u 0 -v $(pwd):/work x1plusbuild make
> > ```
> 
> Hmm. Maybe, yeah. I'll do some testing.

Sure, what I did is just to force use UID 0 which is the root user in the container.

PR #361, opened 2024-09-08
x1plusd.expansion: add GPIO output controls

jwise:
In the X1Plus Expansion Board world, we suddenly grow the concept of external GPIOs.  (Some day soon, we will also have internal GPIOs -- the buttons on the top of the printer!)  The X1P-005 Andon Board has a GPIO output that one might wish to trigger from Gcode -- it has a beeper.  The X1P-006 Shutter Release Board also has two useful GPIOs that one might wish to use to, er, well, trigger camera shutter releases.

In this PR, we add a mechanism by which parts of the system can define GPIOs, and provide an abstraction layer to access them.  The two problems that we try to solve are that not all GPIOs are the same, and that GPIOs can have many semantic descriptors.  The first of those, indeed, is that not all GPIOs are created equal, even as FTDI GPIOs: GPIOs on a SPI-opened FTDI device behave differently from GPIOs on an I2C-opened FTDI device, which behave differently from GPIOs on a FTDI device that has no special function otherwise.  The GPIO API provides an abstract base class, `Gpio`, that a GPIO provider needs to override.

The second of these is that users may want to specify GPIOs in many different ways.  For instance, a user might want to trigger all camera shutters on the system at once, without caring which one they're after.  Or they might want to trigger both camera shutters on a single X1P-006, but not camera shutters on another X1P-006 in a different slot.  Or they might want to trigger a specific pin that doesn't have a functionality named for it.  The GPIO API solves this by naming GPIOs as dictionaries of attributes, and providing some implicit attributes for each port, as appropriate.  For instance, a user might wish to define an X1P-005 board as having a GPIO with dictionary `{"pin": 3, "function": "buzzer"}`, but the GPIO also gets additional implicit properties to describe the board and port it is attached to: `{'port': 'a', 'board_type': 'X1P-005', 'board_serial': 'X1P-005-B01-00000001', 'pin': 3, 'function': 'buzzer'}`

A user can select a GPIO by filtering by one or more of its attributes.  For instance, that GPIO could be selected as `{"port": "a", 'pin": 3}`; as `{"function": "buzzer"}`; or even as `{"board_serial": "X1P-005-B01-00000001"}`.

This PR defines an initial provider for GPIOs from the `ledstrip` expansion driver, but future work would be to have a raw `gpio` expansion driver -- and, of course, to have a provider for the input button GPIOs on the system, and the chamber and toolhead lights as output GPIOs.  This PR also defines a simple client for GPIOs as a `gpio` Action: `{"gpio": {"action": "pulse", "duration": 0.2, "gpio": { "function": "buzzer" }}}`.  This PR does not support GPIO input, but this would be useful in the future as well.

To test this, consider attaching an X1P-006-B01 to an expansion board, and setting the `expansion.port_a` setting to the value `{"ledstrip": {"leds": 9, "gpios": [{"pin": 3, "function": "buzzer"}]}}`.  Then, try Gcode of the form:

```
M400
;x1plus define 2 [{"gpio": {"action": "pulse", "duration": 0.2, "gpio": { "function": "buzzer" }}}, {"delay": 0.2}, {"gpio": {"action": "pulse", "duration": 0.2, "gpio": { "function": "buzzer" }}}, {"gcode": "M400 W0"}]
M73 P50 R0
M976 S99 P2 
M400 W1
```

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
@@ -34,7 +34,7 @@ class X1PlusDaemon:
         self.gpios = GpioManager(daemon=self)
         self.expansion = ExpansionManager(router=self.router, daemon=self)
         self.mcproto = MCProtoParser(daemon=self)
-        self.actions = ActionHandler(daemon=self)
+        self.actions = ActionHandler(router=self.router, daemon=self)
         if not self.settings.get("polar_cloud", False):
             self.polar_cloud = None
         else:
@@ -51,6 +51,7 @@ class X1PlusDaemon:
         asyncio.create_task(self.sensors.task())
         asyncio.create_task(self.expansion.task())
         asyncio.create_task(self.mcproto.task())
+        asyncio.create_task(self.actions.task())
         if self.polar_cloud:
             asyncio.create_task(self.polar_cloud.begin())
 
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
@@ -13,6 +13,7 @@ from .expansion import ExpansionManager
 from .sensors import SensorsService
 from .mcproto import MCProtoParser
 from .actions import ActionHandler
+from .gpios import GpioManager
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


riptidewave93 (approved):

bani6809 (approved):
seems to work

jphannifan (approved):
I still haven't gotten around to testing this with real GPIOs yet (busy week) but it does work on my printer with no issues at all. 

I like how `GpioManager` is structured and especially with PIO integration we should now eventually be able to do some cool things like 

- PWM pins and possibly even a "high speed" pin mode
- PIO programs mapped to GPIO states - with the buzzer example this could look something like this:
```
@rp2.asm_pio(set_init=rp2.PIO.OUT_LOW)
def buzzer_program():
    wrap_target()
    set(pins, 1)
    set(x, 31)
    label("delay_on")
    nop()
    jmp(x_dec, "delay_on")
    set(pins, 0)
    set(x, 31)
    label("delay_off")
    nop()
    jmp(x_dec, "delay_off")
    wrap()
```

We would need to implement PIO operations in the GPIO abstraction layer first, but I think that's beyond the scope of this PR. This looks good for a first cut.

PR #362, opened 2024-09-10
Add build script for clientside installer

shitwolfymakes:
~~This PR updates the build instructions for the client side installer. It is tested and working for me, but more confirmation is appreciated.~~

I wrote a whole build script, I'm too lazy to copy and paste from a README. I did update the README to account for the changes I made. It's tested and working.


jphannifan (approved):
Thanks for this! This is long overdue. 

It looks good to me, and the only things I would change here are minor nitpicks:

- specify that this will only run on macOS or Linux - windows users can install WSL 2 and eventually be able to run this, but that's probably not obvious to beginners
- either move `pack-em-all.sh` up a directory or move `build.sh` down a directory. It's slightly confusing having the two scripts in separate folders

shitwolfymakes:
I actually strongly disagree on moving `pack-em-all.sh`. That script is building and packing each distribution of the client-side installer, and indeed `clientside-installer/install-gui/` is actually the root folder for the installer electron project, NOT `clientside-installer/`.

I have updated the README as requested, good catch!

Issue #107, opened 2024-03-19
Option to wipe SD card during installation process.


RobinWhitfield:
**Is your feature request related to a problem? Please describe.**
Currently the printer will fail to boot if the partition setup is not correct (see #81), and completely wiping all partitions off an SD card is not always simple or obvious (for example on Windows if there's a filesystem it doesn't recognise).

**Describe the solution you'd like**
Add an option within the installer to completely wipe the micro SD card before installing X1Plus.

**Software Information**
 - Current X1Plus Installer: 1.0

bani6809:
this will have to be done via installer-clientside before the x1p is copied over. i tried to do code to wipe the sdcard printer side but you can't do that as the sdcard is already in use by that time.

3DJupp:
But another question would be the disk size:
X1Plus requries **FAT32**, which is limited to **32GB**, at least on most operating systems.
This is not yet included in the [official documentation](https://github.com/X1Plus/X1Plus/wiki/Installation-Guide#getting-started)

> A MicroSD card, **16GB or larger**, FAT32 formatted. High Endurance or Industrial cards are recommended.

jwise:
We should now refuse to install with an unsupported partition configuration.  Actually wiping it during the install is kind of tricky.  What do you think about the resulting state of affairs, @RobinWhitfield ?

bani6809:
installer-clientside could ssh in and have a look around before even starting.

shitwolfymakes:
I'll be implementing this soon

PR #363, opened 2024-09-11
Add SD card validation step to client-side installer

shitwolfymakes:
This PR add a new step to the client-side installer. The new step does the following:

1. Checks the filesystem on the SD card to ensure it is already formatted as FAT32. If not, it stops the installation process and notifies the user.
2. Checks the available space on the SD card to ensure there is enough room to install X1+. The default install size is ~1.5Gb, so I set the check for >2GB to leave room to grow, and an easy way to modify that check in the code later on. If there is not enough room on the SD card, it stops the installation process and notifies the user.

This PR could be extended to implement #107, but if that is a no-go then that issue can be closed as rejected with the merge of this PR in its current state.

I have tested the new installer and confirmed that it's working.

### DIFF
--- a/installer-clientside/build.sh
+++ b/installer-clientside/build.sh
@@ -1,0 +1,27 @@
+#! /bin/bash
+
+### THIS ONLY WORKS IF THE USER HAS BEEN ADDED TO THE DOCKER GROUP
+### RUNNING THIS WITH `sudo ./build.sh` RESULTS IN DUBIOUS OWNERSHIP
+### ERRORS WHEN BUILDING THE FIRMWARE
+
+cd ..
+# check if x1plusbuild docker image exists
+if ! docker image ls --format '{{.Repository}}' | grep -q '^x1plusbuild$'; then
+    echo "x1plusbuild docker image not found, building..."
+    docker build -t x1plusbuild scripts/docker/ || exit 1
+fi
+
+# build the installer
+echo "Building the electron app..."
+docker run -u "$(id -u)" -v "$(pwd)":/work x1plusbuild bash -c 'cd installer-clientside/x1p-js && npm i && npm run build' || exit 1
+
+# build the firmware
+echo "Building the latest firmware..."
+[ -L latest.x1p ] && rm latest.x1p # delete the symlink to the old firmware before building the new one
+docker run -u "$(id -u)" -v "$(pwd)":/work x1plusbuild bash -c 'make && ln -s "$(jq -r '.cfwVersion' ota.json).x1p" latest.x1p' || exit 1
+
+# package the installer with the latest build of the firmware
+echo "Packaging everything..."
+docker run -u "$(id -u)" -v "$(pwd)":/work x1plusbuild bash -c 'cd installer-clientside/install-gui && npm i && bash pack-em-all.sh' || exit 1
+
+cd installer-clientside || exit 1
--- a/installer-clientside/install-gui/src/index.ts
+++ b/installer-clientside/install-gui/src/index.ts
@@ -410,6 +410,32 @@ const legacyInstallSteps: InstallStep[] = [
 
 const firmwareRInstallSteps: InstallStep[] = [
   {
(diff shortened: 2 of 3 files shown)


installer-clientside/install-gui/src/index.ts
```diff
@@ -409,6 +409,32 @@ const legacyInstallSteps: InstallStep[] = [
...
+  {
+    label: "SD card validation",
+    fn: async () => {
+      // Get the format of the sd card (expecting FAT32)
+      console.log("install: checking sd card format...");
+      props.intraStatus = "Checking SD card format...";
+      let result = await printer.sshClient.execCommand("SDCARDDEV=$(df -Th /mnt/sdcard | awk 'NR==2 {print $1}') ; fsck.vfat -n $SDCARDDEV | grep -q FAT32");
+      if (result.code != 0) {
+        console.log("SD card is NOT FAT32 formatted!");
+        throw `SD card is NOT FAT32 formatted!`;
```
bani6809, line 421: we should not just error, we should tell the user what to do. use the same language as installer/install.py

shitwolfymakes: Good catch, I've fixed this in a4a0db9

installer-clientside/build.sh
```diff
@@ -0,0 +1,27 @@
+#! /bin/bash
+
+### THIS ONLY WORKS IF THE USER HAS BEEN ADDED TO THE DOCKER GROUP
+### RUNNING THIS WITH `sudo ./build.sh` RESULTS IN DUBIOUS OWNERSHIP
```
riptidewave93, line 4: Maybe check for root, and error out so we don't build?

```
if [ "$EUID" -eq 0 ]; then 
  echo "Please do not run as root/with sudo. Exiting..."
  exit 1
fi
```

riptidewave93 (approved):

riptidewave93:
LGTM but one small bit of feedback

installer-clientside/build.sh
shitwolfymakes: Great idea! Implemented with c3ab0d7

shitwolfymakes:
@jwise ready to merge unless you want me to add formatting as a hidden advanced object while I'm working on this

installer-clientside/install-gui/src/index.ts
jwise: that commit does not appear to be on the branch to pull?

jwise: fsck.vfat is not really the way to do this (`blkid` probably is fine), but more to the point, this does not actually check the same thing as `install.py`, nor does it check the partition table; this check could pass and `install.py` could still reject the SD card.

relatedly, this seems like the sort of thing that we could detect before tripping on the installer?  it seems in particular like the sort of thing that could error in `connectSsh`, before the user even gets as far as clicking install and then having to exit and restart the installer to recover...

shitwolfymakes: > that commit does not appear to be on the branch to pull?

My bad, committed it to a different branch on accident. Good catch! Fixed with b01b9e4

shitwolfymakes: > fsck.vfat is not really the way to do this (`blkid` probably is fine), but more to the point, this does not actually check the same thing as `install.py`, nor does it check the partition table; this check could pass and `install.py` could still reject the SD card.
> 
> relatedly, this seems like the sort of thing that we could detect before tripping on the installer? it seems in particular like the sort of thing that could error in `connectSsh`, before the user even gets as far as clicking install and then having to exit and restart the installer to recover...

The check here is basically identical? `validate_sd()` checks if the filesystem is vfat. The command I wrote goes a step further and confirms that the fs is FAT32, as vfat applies to FAT32, FAT16, etc.

We could check earlier in the install process, but adding it here offered the cleanest drop-in of new code, without affecting anything else.

Issue #368, opened 2024-09-21
expander: GPIO inputs


jwise:
there should be GPIO input actions -- for now, GPIO wait-for actions would be fine.

jwise:
I think this probably should be implemented, actually, in terms of an API from #369.

PR #376, opened 2024-09-30
x1plusd: GPIO input support

### LINKED ISSUE #369: x1plusd: GPIO toggle / interrupt support
GPIOs should be able to report interrupts, and should be able to have actions associated with them.

in the same breath, it would be good to have the gpiokeys turn to x1plusd generic GPIOs, and for them to trigger actions also -- and for the current gpiokeys actions to turn into x1plusd actions like any other (so they can be triggered from expansion board GPIOs)

jwise:
Adds support in the `GpioManager` for various features that are necessary to take input from system GPIOs.  Specifically:

* Adds an `on_event` mechanism that can allow an internal x1plusd component to temporarily or permanently register a callback for any matching set of GPIOs, or to get a queue of events from the GPIO.
* Adds support for polling GPIOs from sources that do not support interrupts (like the FTDI).
* Implements the `gpio` action's `wait` command using a temporary `on_event` receiver that lasts for the life of the `gpio` action.
* Implements settings-backed GPIO actions.

You can try some of this with:
```
# x1plus settings set expansion.port_a --json '{"ledstrip": {"leds": 25, "gpios": [{"pin": 3, "function": "buzzer", "default": 0}, {"pin": 5, "default": 1}, {"pin": 6, "function": "button", "inverted": true}, {"pin": 7, "function": "button", "inverted": true}]}}'
# dbus-send --system --print-reply --dest=x1plus.x1plusd /x1plus/actions x1plus.actions.Execute string:'[{"gpio": {"action": "pulse", "gpio": {"function": "buzzer"}, "duration": 0.01}}, {"gpio": {"action": "wait", "gpio": {"function": "button", "pin": 6}, "value": true, "timeout": 5}}, {"gpio": {"action": "pulse", "gpio": {"function": "buzzer"}, "duration": 0.01}}]'
```
which configures the port to support an X1P-006, and triggers an action that makes the X1P-006 (or any other buzzers on the system) beep once, wait for the left button on the X1P-006 to be pressed (with a 5 second timeout), and then makes the X1P-006 beep again.

You can also try:
```
# x1plus settings set gpio.actions --json '[{"gpio": {"function": "button"}, "event": "rising", "action": {"gpio": {"action": "pulse", "gpio": {"function":  "buzzer"}, "duration": 0.01}}}]'
```
which causes all buzzers on the system to chirp any time any GPIO button on the system is pressed.

Finally, you can also differentiate long-presses from short-presses, with the following actions that can be used to jog the Z-axis using the buttons on an X1P-006:
```
# x1plus settings set gpio.actions --json '[{"gpio": {"board_type": "X1P-005", "pin": 6}, "event": "long_press", "action": {"gcode": "G91\nG0 Z7.5"}}, {"gpio": {"board_type": "X1P-005", "pin": 7}, "event": "long_press", "action": {"gcode": "G91\nG0 Z-7.5"}}, {"gpio": {"board_type": "X1P-005", "pin": 6}, "event": "short_press", "action": {"gcode": "G91\nG0 Z2.5"}}, {"gpio": {"board_type": "X1P-005", "pin": 7}, "event": "short_press", "action": {"gcode": "G91\nG0 Z-2.5"}}]'
```

The whole thing makes me think that we need a way to import JSON and YAML files rather than typing big blobs of JSON in on the command line.  But for now, here it is.

Fixes #369 and #368.

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/ledstrip.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/ledstrip.py
@@ -36,6 +36,8 @@ class LedStripDriver():
         
         self.gpio_dir = 0x03
         self.gpio_out = 0
+        self.gpio_last_read_time = 0
+        self.gpio_last_read_data = None
         
         self.gpio_instances = []
         gpios = self.config.get('gpios', [])
@@ -141,6 +143,11 @@ class LedStripDriver():
 ###############
 
 class LedStripGpio(Gpio):
+    # do not actually submit a read more frequently than this: use the
+    # cached result, if the system is polling multiple GPIOs in one polling
+    # cycle
+    READ_CACHE_INTERVAL_MS = 50
+
     def __init__(self, ledstrip, pin, attr):
         # pin: 1 << x
         self.ledstrip = ledstrip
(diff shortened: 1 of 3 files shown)


riptidewave93 (approved):

jwise:
Yeah let me just push a fix as part of the branch so it can go with this merge

PR #378, opened 2024-10-02
Migrate HardwarePage.qml settings

jphannifan:
The LCD brightness and toolhead LED settings are still set by Devicemanager. This migrates those settings.

New keys:

`leds.toolhead` 

`lcd.brightness`

### DIFF
--- a/bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
+++ b/bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
@@ -19,14 +19,10 @@ Rectangle {
     property var fan: [qsTr("Not installed"), qsTr("Installed")]
     color: Colors.gray_700
     property var maintain: DeviceManager.maintain
-    property var brightness: DeviceManager.getSetting("cfw_brightness", 100.0)
-    property var toolheadLED: DeviceManager.getSetting("cfw_toolhead_led", false)
+    property var brightness: X1Plus.Settings.get("lcd.brightness",100)
+    property var toolheadLED: X1Plus.Settings.get("leds.toolhead", false)
+    property var gcode: X1Plus.GcodeGenerator
 
-    Timer {
-        id: dispBrightnessChangeTimer
-        interval: 300; running: false; repeat: false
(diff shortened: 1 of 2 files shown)


riptidewave93 (approved):

bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
```diff
@@ -413,7 +409,11 @@ Rectangle {
             onMoved: {
                 brightness = value;
                 X1PlusNative.updateBacklight(value);
-                dispBrightnessChangeTimer.restart();
+            }
+            onPressedChanged: {
+                if (!pressed) {
+                    X1Plus.Settings.put("lcd.brightness", brightness);
+                }
```
jwise, line 416: this is a dramatically better way to do this, nice find!

bbl_screen-patch/patches/printerui/qml/settings/HardwarePage.qml
```diff
@@ -446,11 +446,11 @@ Rectangle {
             id: toolheadLedButton
             anchors.verticalCenter: toolheadLedTxt.verticalCenter
             anchors.right: buttonGrid.right
-            dynamicChecked: DeviceManager.getSetting("cfw_toolhead_led", false)
+            dynamicChecked: X1Plus.Settings.get("leds.toolhead",false)
             onToggled: {
                 dynamicChecked = checked
```
jwise, line 451: I think this line can die.  then, if it does, magically `x1plus settings put leds.toolhead` will update in the UI.  hmm, but it still wouldn't result in a toggled event...

anyway either way this would not be a regression so definitely +1 from me

jwise (approved):
this looks good to me, I have not tried it on my printer but I am solidly on board with killifying these remaining two cfw_ props in the GUI

bani6809 (changes requested):
does not work. backlight does not save settings. boots up with min brightness always, despite slider showing max brightness

jphannifan:
> does not work. backlight does not save settings. boots up with min brightness always, despite slider showing max brightness

ty for spotting this. I will have a fix ready to test soon

Issue #354, opened 2024-08-03
AMS and AMS hub updates not possible


Alkerion:
**Describe the bug**
Can't update AMS and AMS hub to latest version, the update menu told me I've the latest versions, that's wrong as newest versions exists.

**To Reproduce**
1. go to "General tab" under Options
2. Tapon "Firmware version"
3. Tap on "AMS hub" or "AMS #x"
4. Update not available

**Expected behavior**
Been able to update AMS and AMS hub

**Additional Details:**
 - Printer type (X1/X1C/X1E): X1C
 - X1Plus version: 2.0

AceAsket:
same problem

Alkerion:
Any chance to have an answer regarding this issue ?

jwise:
What version do you expect to be able to update to, and what version are you currently running?

Alkerion:
I expect to update to latest FW relases from Bambu :
https://wiki.bambulab.com/en/x1/manual/X1-X1C-AMS-firmware-release-history

My Fw are :
![20240814_073341](https://github.com/user-attachments/assets/7a9ab93d-bc15-4058-ad67-411e26e5fd6e)

And I've these msg when I want to update :
![20240814_073350](https://github.com/user-attachments/assets/470fcae0-7efd-44d1-9eea-1bb02ac0d982)
![20240814_073356](https://github.com/user-attachments/assets/902f0870-61a6-4bea-b2c8-c162dc5eab56)

ScottJD:
I have not installed Plus yet. But form what I understand you can power down, remove the SD card and power back up to stock firmware and should be able to update the AMS like that. Or if you don't have/allow internet connection to the printer you can follow the OTA offline SD card update procedure form Bambu Labs to update the AMS by putting the AMS firmware on a different SD card and update the AMS that way. I am not saying it’s not a bug, but for those that find this important for the RFID’s of new filaments this is a work around for now. 
https://wiki.bambulab.com/en/x1/manual/X1-firmware-update-from-SD-card

lightmaster:
> I have not installed Plus yet. But form what I understand you can power down, remove the SD card and power back up to stock firmware and should be able to update the AMS like that. Or if you don't have/allow internet connection to the printer you can follow the OTA offline SD card update procedure form Bambu Labs to update the AMS by putting the AMS firmware on a different SD card and update the AMS that way.

[...]

PR #379, opened 2024-10-02
Update base.json for AMS firmware

jphannifan:
This addresses #354 so that users can finally update their AMS through X1Plus' UI. 

There are still issues to resolve such as

a. Hub updates

b. Filament updates


riptidewave93 (approved):

jwise (approved):
Oh I missed this one, sorry about that!  Yes this is good if it works.  We should potentially also add support to update AHB -- and filament.zip -- someday, but for now, this is good.

riptidewave93:
@jphannifan can we get this merged in plz? :)

PR #390, opened 2024-10-27
x1plusd: pmsa003i driver error correction

WolfwithSword:
Adds error correction to the concentration values of the pmsa003i i2c driver.

Sensor values, especially the >0.3 concentration value, can sometimes exceed the max unsigned int (16bit) of 65535. This can be detected if the value of >0.5 is larger than >0.3, and such a rule works for all pairs upward.

This was exhibited when printing with PETG-CF, and reproduced manually by holding a soldering iron with a plastic-welding tip near the sensor, with plastic such that it was smoking a bit. 

To address this, the pmsa003i config can now accept a new parameter "overflow_mitigation", which is false by default.

If true, for each concentration value detected to be less than the one greater (except 10.0, as nothing is greater, but it is exceedingly rare for this to be above the max 16bit int), it will add 65535 to the value.

If false, it will set the value to -1, indicating an overflow error, and additional INFO level logging will be supplied for the respective concentration value.

Additionally, the current setting for overflow_mitigation is added to the published results, so the end-user can know if they can expect -1 or >65535 when a value overflows. This way, output is always of the same type, but can clearly indicate what has happened to the data in either case.

Ex settings:

`x1plus settings set expansion.port_b --json '{ "i2c":  { "0x12": { "pmsa003i": {"name": "particulate_matter", "interval_ms": 30000, "overflow_mitigation": true } } } }'`

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/i2c.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/i2c.py
@@ -163,7 +163,7 @@ class Pmsa003iDriver():
         
         self.interval_ms = int(config.get('interval_ms', 1000))
         self.name = config.get('name', f"{i2c_driver.ftdi_path}/i2c/0x{address:02x}/{self.device_type}")
-        self.error_correction = config.get('error_correction', False)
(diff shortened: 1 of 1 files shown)


jwise (approved):
LGTM.  Nice work!

PR #392, opened 2024-10-27
x1plusd: Dynamic loading of i2c drivers

### LINKED ISSUE #386: Dynamically load/register i2c drivers

**Software Information**

**Additional context**

WolfwithSword:
An initial implementation for dynamic loading of i2c drivers, as well as separating them out from i2c.py.

Rather than implementing a complete registration/deregistration service, this takes advantage of the i2c interface being reloaded when the config for it is changed.

### Steps

- A new directory is created to house each driver file, which *must* contain the `device_type` attribute.

- On i2c init (on boot and on each expansion port config update), it will scan this directory for all py files, load initially each module/class and check if it has the device_type attribute, if so, it adds that to the known DEVICE_DRIVERS. 

- Then when parsing the i2c config, it will only load a device and its associated driver if one matches a known device_type. If none are found, an error log is produced stating no compatible drivers are found.

This works quite well given that updating the i2c config reloads it all anyways, this simply hooks into that behaviour.

I expect quite a few changes necessary, one thing I am looking for is a better method of driver class validation. Something that would also verify the format is as expected.

Example screenshot of logs after reloading the config after adding new driver files and changing device_types in config and files to match/mismatch (was messing with SHT41 driver).

image

Closes #386

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/i2c.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/i2c.py
@@ -1,5 +1,7 @@
 import os
 import logging
+import importlib.util
+import inspect
 
 import pyftdi.i2c
 import binascii
@@ -11,6 +13,7 @@ logger = logging.getLogger(__name__)
 
 class I2cDriver():
     DEVICE_DRIVERS = {}
+    DRIVER_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), './i2c_drivers')
     
     def __init__(self, daemon, config, ftdi_path, port_name):
         self.ftdi_path = ftdi_path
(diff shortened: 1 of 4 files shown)


images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/i2c.py
```diff
@@ -21,13 +24,32 @@ def __init__(self, daemon, config, ftdi_path, port_name):
...
+            if filename.endswith('.py'):
+                module_name = filename[:-3]
+                driver_path = os.path.join(self.DRIVER_DIR, filename)
+                spec = importlib.util.spec_from_file_location(module_name, driver_path)
+                module = importlib.util.module_from_spec(spec)
+                spec.loader.exec_module(module)
+                
+                for name, driver in inspect.getmembers(module, inspect.isclass):
+                    if hasattr(driver, 'device_type'): 
+                        self.DEVICE_DRIVERS[driver.device_type] = driver
```
WolfwithSword, line 40: TODO later - after loading all drivers, and mapping all devices, could delete unused drivers from DEVICE_DRIVERS to remove the class references from memory. Unless this does not seem necessary?

WolfwithSword:
Closing PR for now - still thinking of different approaches that can expand to general x1plusd modules, and needing to find time to do so.

Good ideas mentioned is to, for module loading, have metadata in a block docstring rather than eval'ing each. As well as determining load failure logic. When a new x1plusd module is loaded, could restart all of x1plusd - if crash, unload latest added module?

i2c drivers ideally should still be made into separate files for clarity, but whether or not they should be dynamically loaded each when requested by i2c config, or load *all* regardless whenever the i2c expansion module is reloaded, etc.

PR #400, opened 2024-11-19
x1plusd: Dynamic module and i2c loading

### LINKED ISSUE #386: Dynamically load/register i2c drivers

### LINKED ISSUE #399: Dynamic X1Plusd module loading
Feature request initially sparked by #386 to support at least an initial implementation of dynamically loaded modules for x1plusd

Benefit is to help separate the code logically and be able to enable/disable modules dynamically. Additionally, starts the framework for making it easier for development of custom modules.

X1plusd initializes, scans directory for files/directories for modules (via docstring details) to resolve, import, and load.

Hardcoding hardships

**Software Information**
 - Current X1Plus Version: 2.0

**Additional context**
N/A

WolfwithSword:
Expanding on the WIP i2c loading, I started working on dynamic x1plusd driver loading.

- x1plusd will always require the settings module and dbus module, and are not dynamically loaded
- Module files will be either within `x1plusd` package or a single subpackage folder within the same directory.
- Each module file has a definition block in the docstring which provides the module name, class to import, if it requires the router, key in settings to enable/disable the module loading, default import state if no setting found, required module dependencies, and start function to run as an asyncio task.
- Modules are imported with their respective package name, so relative imports and logging paths still work as expected
- If a module fails to import, only it will fail. During a fail, it won't be added to the daemon attributes, so any modules that depend on it will also not load.
- On module failure, the settings key is still watched for changes.
- Any change to watched settings keys will trigger a full reload of x1plusd

Benefits are being able to choose which modules to load if some are not in use, and a consistent loading strategy that custom modules can match.

Additionally, I kept in the work for dynamically loading i2c drivers, but moved the validation also to a block in the docstring.

Module definition
```ini
[x1plusd-module]
name=name, required
class_name=ClassName, required. Class to load and import
requires_router=bool, default: false, does the class require the router to instantiate
settings_key=settings.path.for.bool, required, setting key to listen to for loading
start_func=function_name, default: task, set as 'null' if it does not have a start function. Runs as asyncio task
required_modules=module1,module2, default: none. Won't instantiate until all requirements are instanced
default_state=bool, default: false. Determines its default enabled state when not present in settings
[end]
```
i2c driver definition
```ini
[i2c-driver]
name=name, required, must match the driver name in i2c configs
class_name=ClassName, required. Class to load and import
[end]
```

Things to do:

- Looking for suggestions for better naming and format of docstring blocks and other vars/logging. Especially on the implementation of settings keys for enabling module loading since it's configurable.
- Custom module directory: Created `custom` folder/package in x1plusd. We could symlink this to a spot on the SD card (also need to create __init__.py?). Untested, but setup exists for it.
  -  Could be separate PR/issue
- Currently, I loop through the detected modules and adding ones with requirements satisfied until a max iteration count. Very basic but not ideal, could be improved with a proper ordered dependency list/map without a loop.
  - Could be separate PR/issue

~~Currently keeping this as draft as I test for about a week, and due to large changes to x1plusd, would be great to have a secondary test as well.~~

Have tested for a week while printing and toggling on/off various modules, while testing stuff such as both i2c and the andon module. Seems stable.

Closes #386
Closes #399

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
@@ -10,6 +10,16 @@ import x1plus.utils
 from .dbus import *
 from .settings import SettingsService
 
+from .ota import OTAService
+from .sshd import SSHService
+from .httpd import HTTPService
+from .mqtt import MQTTClient
+from .expansion import ExpansionManager
+from .sensors import SensorsService
+from .mcproto import MCProtoParser
+from .actions import ActionHandler
+from .gpios import GpioManager
+
 from x1plus.utils import module_loader, module_docstring_parser
 
 logger = logging.getLogger(__name__)
@@ -22,8 +32,7 @@ class X1PlusDaemon:
         "name": {
             "definition": {},
             "module_class": Class,
-            "path": "",
-            "loaded": true/false. If unable to load, deleted from modules
+            "loaded": true/false. False if error loading.
         }
     } 
     """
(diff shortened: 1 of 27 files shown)


WolfwithSword:
Added modularized LED Animation loading and moved similar module loader + docstring setting mapper to utils

WolfwithSword:
Added modularized action registering (on init of actions.py). No class loading needed as the decorators run at import. It just needs the valid docstring segment.

```ini
[action-type]
name=gcode
[end]
```

For future thought - all of these modular loading areas can easily be expanded/changed to search in another custom directory that is not a core directory, reusing the same logic, which is nice.

WolfwithSword:
Modularized I think the last thing needed - Expansion drivers (i2c and ledstrip). 

I think right now things are a bit messy with the core/shipped modules and drivers packaging if we are going to dynamically load things. A restructure and associated code changes for directory lookups and such might be a good idea.

Definition for expansion drivers:

```ini
[expansion-driver]
name=i2c
class_name=I2cDriver
[end]
```

WolfwithSword:
Closing due to new branch in x1plus repo

PR #403, opened 2024-11-29
x1plusd: Dynamic module and driver loading

### LINKED ISSUE #386: Dynamically load/register i2c drivers

### LINKED ISSUE #399: Dynamic X1Plusd module loading

WolfwithSword:
Expanding on the WIP i2c loading, I started working on dynamic x1plusd driver/module loading.

- All core x1plusd modules are untouched
- Custom modules (such as polar_cloud) will be dynamically loaded
  - X1Plusd Module files will be either within `x1plusd` package or a single subpackage folder within the same directory.
  - i2c drivers, led animations, actions have their own directories to load from. Expansion drivers load from the same directory as themselves. These are mostly for organization
  - They must confirm to a docstring standard to be scanned and loaded
- Modules are imported with their respective package name, so relative imports and logging paths still work as expected
- If a module fails to import, only it should fail. If it has a start function, it should perform error handling.
- Each module, whether failed/success or enabled/disabled, will have a settings key `x1plusd.modules.name` true/false, and on change, will reload all of x1plusd.
- SD Card imports can come from `/mnt/sdcard/x1plus/modules`, which will be treated the same as `x1plusd/modules`, but will load in using that package path.

### Single Module-type definition
```ini
[module]
name=name, optional (to override name derived from file name)
enabled=boolean, optional. True = Is enabled if settings key is missing. False by default.
[end]
```
*Note, you can have an empty module definition, even if it looks ugly. It just needs the [module] and [end] to be verified.

With this single module definition, most modules are loaded with the decorator registration system (see `led_animations`, `actions`, and `i2c_drivers`). Other modules, such as `polar_cloud` or expansion modules (`i2c.py`, `ledstrip.py`) are loaded using a `load` and optional `start` mechanism.

All modules are scanned for a `load` function when imported, which will run if found. When starting async tasks, all modules imported will be scanned for a `start` function, which will run if found. Anything using the decorator registration method is run at import time due to the nature of the decorators.

Modules follow a template for imports, trying path and relative imports, one of which should always succeed regardless of the order in which modules are loaded.

All modules from the SD card folder (`x1plus/modules`) and subfolders are loaded *after* all core and core-dynamic modules, so no dependency issues will be found. Additionally, all modules imported in the sd card folder will be imported into the system modules with the package name as if they were found in `x1plusd/modules`. 

Closes #386
Closes #399


images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +20,192 @@
...
+
+                        try: 
+                            if module_data.get("start_func", "") == "null":
+                                if not callable(module_class):
+                                    continue
+
+                                if module_data.get("requires_router", "false").lower() == "false":
+                                    self.custom_modules[module_data.get("name")] = module_class(daemon=self)
+                                else:
+                                    self.custom_modules[module_data.get("name")] = module_class(router=self.router, daemon=self)
```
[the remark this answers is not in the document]
WolfwithSword: Yeah agreed, wasn't quite sure how "strong" the need was to not call it from the daemon given the other examples - I think I will also make the current core-modules consistent then as well to remove confusion.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +20,192 @@
...
+                        if BASE_MODULE_DIR in directory:
+                            root_package = "x1plus.services.x1plusd"
+                            nested_dir = directory.replace(BASE_MODULE_DIR, "").replace("/", ".")
+                            if nested_dir: 
+                                # One level deep, i.e., 'custom'
+                                root_package = ".".join([root_package, nested_dir])
+                                root_package = root_package.replace("..", ".")
+                            package = root_package
+                        if not package:
+                            continue
```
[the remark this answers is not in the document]
WolfwithSword: I agree the formation for especially nested directories/packages is pretty brittle, but this has some benefits. For example, the logger will maintain this package path with this method - a huge improvement over my initial implementation that didn't take into consideration the package at all, and I had to pass in the logger and it either had the wrong package name, or I'd need to overwrite it within each module. Without passing in the logger, it logged very strangely in regards to the package.

Although, if we move to a "single" module directory for all types of modules (while also genericizing them), we could simple just load them with that specific package forced in. That way it's easily traceable for debugging still.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +20,192 @@
...
+
+        BASE_MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
+        # Custom module folder could be in x1plus.services.x1plusd.custom, and symlinked to sd card?
+
+        directories = [BASE_MODULE_DIR]
+
+        for filename in os.listdir(BASE_MODULE_DIR):
+            full_path = os.path.join(BASE_MODULE_DIR, filename)
+            if os.path.isdir(full_path) and not filename.startswith("_"):
+                directories.append(full_path)
```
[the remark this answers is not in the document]
WolfwithSword: I did build this initially for there being able to work with any directory, and much prefer a "here's a single directory with everything" approach. That would just require some architectural changes for say, polar_cloud, which I think given from other comments, is probably fine to do now. 

Subdirs could be used for actions/drivers/animations, more so of a means of organization.

As for the sd card path, that can easily be added as a second directory to look into, or we could symlink the base module directory to the sd card. Either one works I think.

WolfwithSword:
Taking review suggestions for a few days to make some broader changes and testing, then will get back with have done's, can do's, and changes if necessary. After the first round of me refactoring and such we can check in again and see if something more concrete is still needed and if so, then either of us can take a crack at the changes.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +20,192 @@
...
+                        [end]
+                        """
+
+                        if not module:
+                            continue
+                        if not hasattr(module, module_data.get("class_name")):
+                            logger.warn(f"Could not load {module_name} in x1plusd module loader. Class not found: {module_data.get("class_name")}")
+                            continue
+
+                        self.settings_keys.append(f"x1plusd.modules.{module_data.get("name")}")
```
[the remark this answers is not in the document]
WolfwithSword: Purely used for listening to changes to restart all of x1plusd. It could have future use such as fetching all settings keys that x1plusd is currently watching for in regards to dynamic loading, whether enabled or disabled. But there are many other ways to do that.

Didn't like my name initially, will change

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/manager.py
```diff
@@ -115,7 +117,37 @@ def __init__(self, daemon, **kwargs):
...
+                if not hasattr(module, driver_data.get("class_name")):
+                    logger.warn(f"Could not load {module_name} in expansion driver loader. Class not found: {driver_data.get("class_name")}")
+                    continue
+
+                try:
+                    driver_class = getattr(module, driver_data.get("class_name"))
+                    self.DRIVERS[driver_data.get("name")] = driver_class
+                    logger.info(f"Loaded Expansion Driver: {driver_data.get("name")}")
+                except Exception as e:
+                    logger.error(f"Failed to load Expansion Driver '{driver_data.get("name")}': {e.__class__.__name__}: {e}")
```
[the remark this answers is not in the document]
WolfwithSword: Going to attempt to move this to the `expansion.__init__` as a secondary module loader with a different directory. This way it should be a bit more scalable. But if it doesn't work out, will revert to builtins for the time being.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/ledstrip_animations/finish.py
```diff
@@ -0,0 +1,33 @@
...
+        self.timeout = config.get('timeout', 600) # success goes away after 10 minutes
+        self.last_finish_trn = 0
+        self.last_was_finish = False
+    
+    def can_render(self):
+        finish = self.leds.last_gcode_state == 'FINISH'
+        if not self.last_was_finish and finish:
+            self.last_finish_trn = time.time()
+        self.last_was_finish = finish
+        return finish and (time.time() - self.last_finish_trn) < self.timeout
```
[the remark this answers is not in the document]
WolfwithSword: I still like the idea of separation as a best practice, however if we might expect a user to load say, a custom "animation library"  which is just a file with more than one animation to register, it should work and we should have an example of this built in.

Agree, will merge all the common state-based ones into a "states.py" or similar, and keep rainbow out.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
```diff
@@ -39,16 +46,21 @@ def __init__(self, daemon):
...
+            # Assign socket callbacks.
+            self.socket.on("registerResponse", self._on_register_response)
+            self.socket.on("keyPair", self._on_keypair_response)
+            self.socket.on("helloResponse", self._on_hello_response)
+            self.socket.on("welcome", self._on_welcome)
+            self.socket.on("delete", self._on_delete)
+        except Exception as e:
+            logger.error(f"Failed to start polar_cloud: {e.__class__.__name__}: {e}")
+            if self.socket:
+                self.socket.disconnect()
```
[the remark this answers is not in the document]
WolfwithSword: Generally yes, and if I force an error elsewise it does work with the isolation. The problem specifically with it here was the socket would not disconnect on teardown and it caused some major issues. Felt it was better for this particular case to do this.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/manager.py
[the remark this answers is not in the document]
WolfwithSword: Ended up treating them just like modules, but they get loaded always after the expansion instance, and in terms of hierarchy - the i2c drivers or led animations can use relative imports within a modules directory for their registration methods, or if erroring, will throw and error and carry on.

My next push will clarify this a bit.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__main__.py
```diff
@@ -30,12 +30,31 @@ def exceptions(loop, ctx):
 
 async def start():
     x1plusd = await X1PlusDaemon.create()
-    await x1plusd.start()
+
+    for key in x1plusd.settings_keys:
+        x1plusd.settings.on(key, lambda: stop())
```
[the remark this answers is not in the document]
WolfwithSword: Kept it outside in `__main__` as that's where the main event loop is held, and that I can just call the method to start tearing it down. I like to think of it as anything that creates, alters, or destroys this main event loop should be in one place.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__main__.py
```diff
@@ -30,12 +30,31 @@ def exceptions(loop, ctx):
...
+
+async def stop_loop():
+    logger.info("Stopping x1plusd after module config change")
+    await asyncio.sleep(2)
+    current_task = asyncio.current_task(loop=loop)
+    tasks = [task for task in asyncio.all_tasks(loop=loop) if task != current_task]
+    for task in tasks:
+        task.cancel()
+    await asyncio.sleep(5)
+    quit()
```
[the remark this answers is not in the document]
WolfwithSword: This is for tearing down the loop based on request, not tearing it down when an error terminates it (and this also ends up running the `finally` block).

WolfwithSword:
Updates:

All types of dynamically loaded modules/plugins are now under a single directory (with subdirectories) `x1plusd/modules` and have a module definition block of just type `module`. 

The only config items recognized at the moment are `name` to override the parsed name if necessary (optional), and `enabled` which is whether or not to have its default state be loaded unless settings are False. 

Most modules are based off a load and register system with a decorator (actions, led_animations, i2c_drivers), however expansion modules are loaded by a load function, and as is all x1plusd custom modules at the root modules directory. Anything imported dynamically with a `load` function will have it run after import is successful.

Anything hierarchy-dependent is resolved based on refactoring of load and start functions, as well as when necessary (such as i2c drivers or animations and their respective expansion module), using relative imports instead of full path - but only to other dynamically loaded imports. If there is ever an error in loading like this, it is simply logged and passed over. See `polar_cloud.py`, `i2c.py`, or `ledstrip.py` for examples, as well as `i2c_drivers` or `led_animations` folders.

Once all modules are loaded, then the async tasks are started for all core modules, and any module with a `start` function in the import (not class) will be run. See `polar_cloud.py` for an example.

Logging added to utils class and utility function added for loading from a provided folder-path with overrides for package and recursive lookups. Currently the loader type `module` is hard-coded. If we ever need to change this (for example, we decide to differentiate between modules and plugins (for hierarchy reasons)) it can be done easily.

Other changes are consistency around passing or using `router` to all core modules.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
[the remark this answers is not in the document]
jwise: one other thing I was thinking about in the shower a few days ago was the concept of distributing multiple python files together as a module.  I did not come up with any good solutions, and I don't think it has to go into this version -- single file load would still be a huge improvement.  but here are some things I was thinking about as possible design constraints for improvements later:

* the general idea is that it would be sort of annoying that current Modules have to be exactly one file -- or otherwise you have to keep track of them separately.  for instance, if the whole expansion subsystem were an X1Plus Module, you would be copying many files to the SD card, and versioning them all independently, but presumably a plugin developer would want to release one file that gets loaded in atomically.
* one thing that we could do is load Python eggs, which are just zip files that can be stuck in a pythonpath, as I understand it.  seems ok, but eggs are deprecated, and more to the point, if there need to be native components with native binaries, then they can't really go into eggs.  and even more likely, it might be the case that at some point we have a mechanism to load qml files into bbl_screen -- and those would want to have qml files in the same directory as py files.  that would not work well with eggs either.
* I wonder, then, if in the long term, X1Plus Modules are something that get installed to /userdata (or to the rootfs overlay, and then wiped out when X1Plus is reinstalled?), and are just zipfiles that the front-end unpacks?

anyway, these are things to think about.  I don't like any of these answers yet, and since I don't like anything particularly better than what you propose so far, I'm definitely ok with continuing down this path as at least one option!  and anyway, single-file python Modules should always be supported, since they are a lightweight way to get stuff running.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/polar_cloud.py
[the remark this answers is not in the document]
jwise: the polar_cloud thing is going to be merged later by ripping the whole thing apart I think, so I'm OK with this, anyway

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +18,94 @@
...
+        self.gpios = GpioManager(daemon=self)
+        self.actions = ActionHandler(daemon=self)
         
+        self.custom_modules = {}
+        """
+        { "name": ClassInstance }
+        """
+
+        self.watched_keys = []
+        self.MODULES = []
```
jwise, line 54: I am not a Python expert but ALL_CAPS is I think traditionally used for global configuration / constant state.  I am not really sure what the actual convention is, but I think it is definitely not for a class member.  I *think* the actual name for this is, like, `self.modules_available`?

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +18,94 @@
...
+        self.expansion = ExpansionManager(daemon=self)
+        self.gpios = GpioManager(daemon=self)
+        self.actions = ActionHandler(daemon=self)
         
+        self.custom_modules = {}
+        """
+        { "name": ClassInstance }
+        """
+
+        self.watched_keys = []
```
jwise, line 53: the interplay between this and the shell script driver seems weird but I'm not sure how to have a better answer than that.  I wonder if a watched key should throw an exception up to the toplevel somehow or something, which would cause th erestart

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +18,94 @@
...
+            os.makedirs(SDCARD_MODULE_DIR)
+
+        # Custom modules are loaded after all core modules
+        # Full import paths will work if a dependency exists
+        self.load_modules(SDCARD_MODULE_DIR)
+
         return self
 
+    def load_modules(self, module_directory: str):
+        self.MODULES.extend(module_importer(module_directory, include_subdirs=True))
```
jwise, line 70: so I think we don't actually want to do the actual `import` until we've agreed (from the settings subsystem) that it's ok to import -- we could haul in a bunch of dependencies (for people trying to run a lean-and-mean x1plusd), and the act of `import`ing something is sufficient to cause code to execute!  so I think `MODULES` ought really be like `modules_available`.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
```diff
@@ -15,44 +18,94 @@
...
         asyncio.create_task(self.ota.task())
         asyncio.create_task(self.httpd.task())
         asyncio.create_task(self.sensors.task())
         asyncio.create_task(self.expansion.task())
         asyncio.create_task(self.mcproto.task())
         asyncio.create_task(self.actions.task())
-        if self.polar_cloud:
-            asyncio.create_task(self.polar_cloud.begin())
+
+        for scanned_module in self.MODULES:
```
jwise, line 100: for that point, perhaps this ought be split into modules_available and modules_enabled -- and then we don't have to go through the settings dance here, we just have to iterate through the modules_enabled

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/actions.py
```diff
@@ -35,11 +35,14 @@
...
 """
 
+import os
 import asyncio
 import logging
 import json
 import yaml
 
+from x1plus.utils import module_loader, module_docstring_parser
+
```
jwise, line 45: dead code

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/mcproto.py
```diff
@@ -263,6 +263,9 @@ async def trigger_action(self, action):
         """
         Trigger an action to execute in the background, with only one MC protocol action allowed to run at a time.
         """
+        if not hasattr(self.daemon, 'actions'):
+            return
+
```
jwise, line 268: dead code

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/httpd.py
```diff
@@ -16,10 +16,10 @@
 PROTOCOL_VERSION = 0
 
 class HTTPService():
-    def __init__(self, router, daemon, **kwargs):
+    def __init__(self, daemon, **kwargs):
         self.daemon = daemon
         self.daemon.settings.on("http.enabled", lambda: asyncio.create_task(self.sync_startstop()))
-        self.router = router
+        self.router = daemon.router
```
jwise, line 22: I'm on board with homogenizing this, nice work

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/actions/delay.py
```diff
@@ -0,0 +1,18 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+@register_action("delay")
+async def _action_delay(handler, subconfig):
+    logger.debug(f"delay action: {subconfig}")
+    if type(subconfig) != int and type(subconfig) != float:
+        raise TypeError(f"delay parameter {subconfig} was not numberish")
+    await asyncio.sleep(subconfig)
```
jwise, line 18: my only thought on this is that it is a good example, but I really would prefer to have core behavior not be modular, lest there be users forgetting they disabled something and then wondering why it does not work

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/actions/syslog.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+from x1plus.services.x1plusd.actions import register_action
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+@register_action("syslog")
+async def _action_syslog(handler, subconfig):
+    logger.info(f"syslog action: {subconfig}")
```
jwise, line 14: *this* is a good sample action, though.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/expansion/i2c.py
```diff
@@ -0,0 +1,76 @@
...
+        for address, devices in config.items():
+            address = int(address, 0)
+            
+            for device, subconfig in devices.items():
+                try:
+                    # Only load device with driver if requested by I2C config
+                    if driver := self.DEVICE_DRIVERS.get(device, None):
+                        self.devices.append(driver(address = address, i2c_driver = self, config = subconfig))
+                    else:
+                        logger.error(f"failed to initialize driver for {device}@0x{address:2x} on {ftdi_path}. Compatible driver not found")
```
jwise, line 48: I think this no longer demand-loads, right?  also should this just look up into `_registered_drivers`?

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/i2c_drivers/aht20_driver.py
```diff
@@ -0,0 +1,91 @@
+"""
+[module]
+name=aht20
+enabled=true
+[end]
+"""
+try:
+    from x1plus.services.x1plusd.modules.expansion.i2c import register_driver
+except ImportError:
+    from ..expansion.i2c import register_driver
```
jwise, line 10: ah, yeah, this is the bit I was worried about :) this means that there needs to be a dependency problem of order in how drivers are loaded.  specifically, this does not necessarily load the i2c that came from a module, but from this on-disk one.  so maybe the expansion core drivers really need to be built-in for now, and the i2c drivers should be loadable?

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/led_animations/states.py
```diff
@@ -0,0 +1,152 @@
+"""
+[module]
+enabled=true
+[end]
+"""
+try:
+    from x1plus.services.x1plusd.modules.expansion.ledstrip import register_animation
+except ImportError:
+    from ..expansion.ledstrip import register_animation
```
jwise, line 9: same concern as i2c

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/led_animations/states.py
```diff
@@ -0,0 +1,152 @@
...
+def _animation_paused(handler):
+    handler.ANIMATIONS.setdefault("paused", PausedAnimation)
+    
+@register_animation("finish")
+def _animation_finish(handler):
+    handler.ANIMATIONS.setdefault("finish", FinishAnimation)
+    
+@register_animation("failed")
+def _animation_failed(handler):
+    handler.ANIMATIONS.setdefault("failed", FailedAnimation)
```
jwise, line 151: hmm, why not just have these be class decorators?

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
```diff
@@ -1,7 +1,24 @@
...
+
+slh = logging.handlers.SysLogHandler(address="/dev/log")
+slh.setLevel(logging.INFO)
+slh.setFormatter(logging.Formatter("%(name)s: %(message)s"))
+logger.addHandler(slh)
+
+ch = logging.StreamHandler()
+ch.setLevel(logging.DEBUG)
+ch.setFormatter(logging.Formatter("[%(asctime)s] %(name)s: %(levelname)s: %(message)s"))
+logger.addHandler(ch)
```
jwise, line 21: or arguably this should just be hoisted into the x1plus package itself to make sure that we get the logger in everything x1plus-ish

jwise (commented):
OK I think this is starting to come together.  Do you mind if I take a swing at making some of the changes that I propose?  Do you object to any of them in particular?

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
WolfwithSword: I think tossing up an exception would theoretically work, though it feels a bit like a bad design if the intent is to force an intentional restart. 

Though this would let us perhaps add the listener and function to throw the exception within `__init__.py` rather than up in `__main__`. I'm okay with either method.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/__init__.py
WolfwithSword: I like this, would also let us poll for it later if we wanted a quick glance as what's enabled or not without needing to iterate and check.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/actions/delay.py
WolfwithSword: I'm 50/50 on it - if a user disabled something like that, it would be a very obvious insert in their `x1plus settings` and be very hidden. And standardization has its benefits especially since it's not wildly unique, hence the compromise I had where this is modular, but it's in the core directories and is always enabled by default.

But I could be convinced if we kept say, some single one modular as an example, and the rest we re-merged into a single file within the `actions.py`. If it's meant to be extremely core, then it doesn't necessarily need to be structured modularly either, though it is very simple/clean to so yeah, 50/50 on it.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/expansion/i2c.py
WolfwithSword: It no longer demand-loads/imports but it demand-instantiates. All the classes are loaded, but it only instantiates the driver on demand from the i2c port config.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/i2c_drivers/aht20_driver.py
WolfwithSword: Is there much risk with this approach though, if it prioritizes one from the import path and as a backup, uses on-disk relative pathing? If it's not loaded into the system modules yet, it uses the on-disk one up until the on-disk one is imported itself into system-modules, bringing everything with it.

The benefit is for when a user has a custom module on the sd-card - it should always end up using the module path first as custom sd-card modules are all loaded *after* they get added to the system modules.

Then to template, these existing modules in core directories have both import options to get around dependency issues as you mention - which only exist for this decorator type function, due to when and where the modules are actually loaded, and for some, started. I was testing this with completely wiped pycache's and it seems to not matter and all still get added to `i2c` or `ledstrip` or `actions`, despite forcing a dependency order issue. As it gets imported from the disk during x1plusd's `__init__` anyways, it seems to be fully functional and brings everything with it.

While not done here, this similar pattern could also be used for custom sd-card modules that have non-module dependencies in other files. It could hope it's on the path, fail, then load local.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/led_animations/states.py
WolfwithSword: Was copy-pasting it since these were at some point separated out into individual files for consistency. They could be, but this way it's slightly more consistent with non-core animations that are loaded, plus was the decorator usage already in place before PR.

WolfwithSword:
> OK I think this is starting to come together. Do you mind if I take a swing at making some of the changes that I propose? Do you object to any of them in particular?

Pushed a quick commit to remove the dead code callouts - thanks!

Feel free to take a crack at any of your suggestions - I agree with most, and left one or two comments on a few others for reasoning/alternatives.

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/i2c_drivers/aht20_driver.py
jwise: Hm, I believe that it worked, but I'm also not sure why it *must* work.  I think my concern is this: if there's one in x1plus in the rootfs, and one on the SD card, then which one you get might not be consistent between modules (i.e., if it was an in-rootfs module or if it was an on-SD-card module) -- otherwise, that `except ImportError:` wouldn't be necessary ever, right?

And, beyond that, even if you do consistently get one loaded from the SD card, it also means that you can end up downgrading something that x1plusd internally depends on across x1plus version upgrades.

I think I am going to haul the i2c and ledstrip drivers back into the core for now.  But we should definitely revisit this in the future.

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
```diff
@@ -37,3 +54,106 @@ def get_IP() -> str:
...
+            modules.append(Module(file_path, package, loader_type = loader_type))
+        except ImportError:
+            # it just had no docstring keys; that's fine
+            pass
+        except Exception as e:
+            logger.error(f"failed to load module metadata from {file_path}: {e.__class__.__name__}: {e}")
+    return modules
+
+    
+def module_docstring_parser(filepath: str, loader_type: str) -> dict:
```
jwise, line 131: back of my mind thought: rather than rolling our own parser, should we just yaml.safe_load?  that way we can have rich contents

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/manager.py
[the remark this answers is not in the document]
WolfwithSword: Resolution was to undo their dynamic loading for now, and export with `__all__` - good work!

images/cfw/opt/x1plus/lib/python/x1plus/utils.py
WolfwithSword: YAML implementation looks nice and is clean/easy to work with, I say it's in the right direction.

jwise (approved):
Seems to work, seems like we're in agreement.  Great work on this!

PR #405, opened 2024-12-02
X1Plus Expander UI

jwise:
Adds a handful of UI to drive the Expander, and a small database of Expander configurations.  LED strips and LED configurations are still to be done, but this is probably a good start for now.  (Realistically, the LED animation system probably needs to change.)

Also left incomplete is having x1plusd actually choose defaults for swapped modules on boot.  This PR is really a MVP to get testers up and running with the board, more than anything else.


bani6809 (approved):
seems to work

riptidewave93 (approved):

images/cfw/opt/x1plus/share/expansion/expansion.json
```diff
@@ -0,0 +1,54 @@
...
+      "driver": "ft2232",
+      "ports": 2,
+      "image": "x1p-002.png"
+    }
+  },
+  "modules": {
+    "generic_i2c": {
+      "name": "I²C",
+      "configuration": {
+        "i2c": []
```
WolfwithSword, line 14: i2c configuration is changed to a list from a dict - should there not be an associated change in i2c.py for a new format, or is this a mistake? Specifically when going over `config.items()` it will fail if the config is stored as a list. 

https://github.com/X1Plus/X1Plus/blob/acf4017d18d907fba0858f9b63dfb6592be29269/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/expansion/i2c.py#L25 

When setting up port_b to generic i2c stemma, it setups the config with the empty list value []. `{"i2c": [], "meta": {"module_config": "generic_i2c"}}`

Manual configuration as dict for config still works, so having the default type here as a list is misleading for configuration unless format changes are needed.

WolfwithSword (changes requested):

images/cfw/opt/x1plus/share/expansion/expansion.json
WolfwithSword: For changes - either we change back to a dict to maintain current functionality, or we update i2c.py to reflect it being a list, and document the new format for configuration of multiple i2c devices.

jwise: I agree, it should be a dict.  Good catch!  Fixed.

PR #417, opened 2024-12-29
x1plusd-i2c: I2C TCA8418 Keypad Driver

WolfwithSword:
Adds i2c driver to core module/driver loading for an i2c keypad driver (tca8418, up to 80 buttons)

TODO: Demo flow of loading in sdcard folder, demo video of button actions, resources for i2c device, examples etc

### DIFF
--- a/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/i2c_drivers/tca8418_driver.py
+++ b/images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/i2c_drivers/tca8418_driver.py
@@ -1,0 +1,214 @@
+"""
+[X1PLUS_MODULE_INFO]
+module:
+  name: i2c.tca8418
+  default_enabled: true
+[END_X1PLUS_MODULE_INFO]
+"""
+
+import logging
+import asyncio
+import time
+
+from x1plus.services.x1plusd.expansion.i2c import register_driver
+
+logger = logging.getLogger(__name__)
+name = "tca8418"
+
+@register_driver(name)
+class Tca8418Driver():
+    """
+    TCA8418 offers up to 8 rows x 10 columns for an up to 80 buttons keypad matrix
+
+    Driver implementation assums user will be starting pinouts at R0, C0, going in order.
+
+    Should a user have a custom or uneven matrix, they should configure the largest number
+    for row/column (as if it's a AxB matrix) and ignore missing key numbers for action-setting.
+
+    Key are numbered 1-80 (inclusive).
+
+    Actions are performed on key-press. MQTT is updated with states of all configured buttons as either `pressed` or `released`
+
+    Configuration Sample:
+
+    { 
+        "tca8418": { 
+            "name": "keypad", 
+            "interval_ms": 100, 
+            "rows": 4, 
+            "columns": 4, 
+            "actions": [
+                {
+                    "key": 1,
+                    "action": {...}
+                },
+                {
+                    "key": 2,
+                    "action": {...}
+                }
+                ...
+            ]
+        }
+    }
+    """
+
+    REG_CFG = 0x01
+    REG_INT_STAT = 0x02
+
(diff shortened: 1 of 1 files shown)


images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/i2c_drivers/tca8418_driver.py
```diff
@@ -0,0 +1,214 @@
...
+                                logger.error(f"Error performing {name.upper()} key-press action for key: {key}, action {action}: {e.__class__.__name__}: {e}")
+                                await self.sensors.publish(self.name, type = name, inop = { 'exception': f"{e.__class__.__name__}: {e}" })   
+                    updated = True
+                  
+                self.write_register(self.REG_INT_STAT, 0xFF)          
+ 
+            except Exception as e:
+                logger.error(f"Error reading {name.upper()} buttons: {e.__class__.__name__}: {e}")
+                await self.sensors.publish(self.name, type = name, inop = { 'exception': f"{e.__class__.__name__}: {e}" })
+                # In case of recurring error, do not spam log/system. Sleep 5s
```
WolfwithSword, line 208: TODO: Fail out of driver entirely if recurrent error - one offs can/should be logged. Recurring errors in loop are rare and would only occur in failure to read, so perhaps if this is hit 5 times in X timeframe (remove sleep) we disconnect driver

jwise:
Some thoughts:

* How about binding the buttons to GPIOs?  I believe that these should be `Gpio`s that have `polling` as `False`, sort of like `MockGpio`.  Maybe their `attrs` are `{ "type": "i2c", "row": y, "col": x, "path": i2c_path, "driver": "tca8418" }` or something?  Or they could have a `poll` functionality that supersedes the `task` that you've otherwise specified -- sort of like how the LED strip GPIO driver works (polling actually looks at all GPIOs, and then if we've done a poll recently for any GPIO on the expander's interface in question, we don't bother to poll again, and we just use the cached state).
* Then, if you do that, you can use the GPIO on-action mechanism: `x1plus settings set gpio.actions --json '[ { "gpio": { "type": "i2c", "row": 0, "col": 0 }, "event": "rising", "action": { ... } }, { "gpio": ..., "event": "rising", "action": ... } ]`
* And if you want to export GPIO actions over MQTT, maybe you could either have an Action to publish to MQTT, or you could have a "publish_mqtt" flag in `gpio.actions` or something?  Or add a `gpio.mqtt_exports` key that works in an analogous fashion?  (Or `gpio.sensor_exports`, for that matter, to use the existing sensors format.)  Then *all* GPIOs on the system could work that way -- including Andon board GPIOs.  (My prior is that we don't want this to be the default for all GPIOs, since polling GPIOs could be expensive, but having this as an option would be good.)

WolfwithSword:
Testing this weekend, but have changed over to gpios interfacing

images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/i2c_drivers/tca8418_driver.py
```diff
@@ -0,0 +1,304 @@
...
+        logger.info(f"probed {name.upper()} sensor at 0x{address:2x}")
+
+        if not self.rows or not self.cols or self.rows > 8 or self.cols > 10:
+            raise Exception(f"{name.upper()} invalid button row/columns configuration")
+
+        self.currently_pressed = set()
+
+        self.initialize_keypad()
+
+        self.name = self.config.get('name', f"{i2c_driver.ftdi_path}/i2c/0x{address:02x}/{name}")
```
WolfwithSword, line 126: Change to i2c_driver.i2c_path after #458

WolfwithSword:
Been testing edge cases on this since my last update and it seems ready for another review round I think.

jwise (approved):
LGTM -- nice work!

PR #424, opened 2025-01-18
docs: update README.md

eltociear:
minor fix


jheyneman:
This is fairly subjective. It depends on if you pronounce the dot (.) or not.

jwise:
I appreciate the PR, but indeed, I do not pronounce the dot :-) I usually say it out loud as "upload an ex one pee file".

PR #430, opened 2025-01-24
Port to 00.00.32.39 (01.08.02.00), and nop out TPU for AMS complaints

jwise:
Changes in 00.00.32.39 are relatively minimal; kick everything forward (and make the netservice shim symbolized to be able to port it to other firmware versions in the future).  While we're in there, nop out the complaints about TPU from bbl_screen, when it's TPU for AMS.

### DIFF
--- a/bbl_screen-patch/interpose.cpp
+++ b/bbl_screen-patch/interpose.cpp
@@ -563,6 +563,10 @@ public:
 
 #ifdef HAS_DBUS
 
+#include "vendor/nlohmann/json.hpp"
+
+using namespace nlohmann;
+
 namespace BDbus {
     class Object;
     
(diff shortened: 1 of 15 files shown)


psiberfunk:
Question : why do we have to NOP out anything , do the updated binaries from bambu not just work as is ?

riptidewave93 (approved):
LGTM but I didn't test.

bbl_screen-patch/interpose.cpp
```diff
@@ -726,6 +730,52 @@ SWIZZLE(void, _ZN5BDbus4NodeC2ERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESa
...
+    } else {
+        if (!proxy) {
+            proxy = dbusListener.nobe->createProxy("x1plus.x1plusd", "/x1plus/settings");
+        }
+        BDbus::Error e;
+        std::string settings = proxy->callMethod("x1plus.settings", "GetSettings", "{}", e);
+        try {
+            json j = json::parse(settings);
+            auto k = j.at("filament.filename").template get<std::string>();
+            printf("get_resource_path: read filement.filename setting from x1plusd of %s\n", k.c_str());
```
errge, line 749: typo here in the error message.

`s/filement/filament/`

jwise: Good catch!  Thanks for the close eyes!

PR #436, opened 2025-02-01
Add Option to toggle light with hardware buttons

phillipthelen:
This is my first PR here, so I hope I modified things correctly.

This is a small change to add an option for "Toggle Light" to the Hardware Settings Page for the physical buttons. This toggles the main/chamber light, the same way the lightbulb button in the touchscreen UI does.

Right now if the touchscreen is active and on the main screen, when the hardware button is used to toggle the light the lightbulb icon on the screen switches between on and of two times until it settles on the correct new state. I'm not sure if this can be fixed/improved, since I haven't worked with qml before. I can however look into it further.


jwise:
Nice work!  Thanks!!

jwise (approved):

PR #484, opened 2025-05-20
expansion: add LED and I2C config support from the GUI

jwise:


bani6809:
"I2C (STEMMA)" pulls up "Port D configuration"

images/cfw/opt/x1plus/share/expansion/I2c.qml
```diff
@@ -1,12 +1,80 @@
...
+ColumnLayout {
+    spacing: 24
+    Layout.alignment: Qt.AlignLeft | Qt.AlignTop
+    Layout.fillWidth: true
+    
+    property var devices_available: [
+        [ 0x44, { "sht41": {} } ],
+        [ 0x34, { "tca8418": { "rows": 4, "cols": 4 } } ],
+        [ 0x38, { "aht20": {} } ],
+        [ 0x12, { "pmsa003i": {} } ],
```
WolfwithSword, line 18: The hex representation is being converted to a string of a decimal, causing confusion/misconfiguration and being unable to pick up currently enabled drivers.

WolfwithSword (commented):
Another bug I cannot determine root cause of based on changed files:

Saving Andon board config for Port A, is somehow triggering my GPIOS I have setup for my i2c tca driver, but only when the i2c tca driver is configured properly.

Trying to look into this but not finding anything immediate.

images/cfw/opt/x1plus/share/expansion/I2c.qml
jwise: Oh yeah, this is a good one.  JavaScript / JSON doesn't have numbers as keys, so we should standardize on using hex as keys.  ba02eb71

bani6809:
filament database updater pops up on every boot, regardless if you install it or not.
image
image
image

WolfwithSword:
Reinstalled new build. It did not recognize the previously UI configured andon module or i2c config, but once saving both once and then reconfiguring i2c via commandline, it picked both up continuously and even identified my in-use i2c modules this time.

Rising events are functional too without being triggered on resave, and still function normally.

Two new issues, but might be a symptom of upgrading - Both the last build I tried and this latest one, upon updating, VNC was disabled and had to be re-enabled via UI. SSH was working, but in the UI showed as disabled. On reboot, they remain enabled. So it's only on update of new x1p. Oh, and video was turned OFF too, but this I've seen in OEM updates pre- 1.08 before.

I don't think those are critical to block RC1, but definitely worth looking at for maybe RC2.

WolfwithSword (approved):
Fixes work, other issues found by others are already noted.

riptidewave93 (approved):

PR #507, opened 2025-07-11
add support for upgrading a secondary AMS

### LINKED ISSUE #496: AMS update fails and hangs UI
**Describe the bug**
Version 3 fails to update ams to version 00.00.06.44 and hangs on second retry.

**To Reproduce**
Go to firmware try to update AMS from 00.00.06.40 to 00.00.06.44
Go thru all the steps to perform the update (it will fail)
then try to update again. it now gets stuck at the updating dont powercycle the printer

**Expected behavior**
Update should either succeed or fail and UI should show a OK / Cancel button depending on outcome.

**Additional Details:**
 - Printer type (X1/X1C/X1E):
 - X1Plus version: 3.0
 - AP firmware version: 00.00.32.39
 - AP hardware revision: AP05
 - MC firmware version: 00.00.27.26
 - MC hardware revision: MC07
 - Toolhead firmware version: 00.00.07.12
 - Toolhead hardware revision: TH09
 - AMS hub firmware version: 00.00.00.42
 - AMS hub hardware revision: AHB01
 - AMS firmware version 00.00.06.40
 - AMS hardware version AMS08

Havent tried, dont want to revert to stock.

**Custom modifications**
-

**Additional context**
-

jwise:
Also, use new AMS FW.

Fixes #496.

### DIFF
--- a/bbl_screen-patch/patches/printerui/qml/settings/UpgradeDialog.qml
+++ b/bbl_screen-patch/patches/printerui/qml/settings/UpgradeDialog.qml
@@ -182,14 +182,20 @@ Item {
             // running!  So we do that.
             X1PlusNative.system(`cp /sdcard/x1plus/firmware/${fileName} /userdata/firmware/${fileName}`);
             X1PlusNative.system(`chmod -x /userdata/firmware/${fileName}`);
-
-            X1Plus.DDS.publish("device/request/upgrade", {
+            
+            var upgrade_pkt = {
                 "command": "start",
                 "sequence_id": "0",
                 "module": module.split("/")[0], /* ams/0 -> ams */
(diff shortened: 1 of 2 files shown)


liftbag:
The AMS units update correctly.
There’s probably just a cosmetic issue: even after the update, the info screen still shows the AMS as if they need to be updated.
However, no update notifications appear when the printer is started.

First boot of the new firmware.
IMG_2982

Screen at machine reboot after AMS update.IMG_2983

liftbag:
> Guh, good catch. That's because I'm a goofball and typoed it as 00.01.06.42 as the version to check against, not 00.01.06.62.

Latest commit works as expected.

IMG_2987

PR #512, opened 2025-07-15
bbl_screen: display sensor data in GUI

### LINKED ISSUE #493: Show sensor readings in bbl_screen UI
An early Expander video showed sensor readings in the bbl_screen UI, but that never quite made the cut (coming up with UI that supported a diversity of sensors wasn't immediately obvious to me, and also, the task slipped my mind because I forgot to create a ticket for it).  Anyway, it would be nice to have a dialog somewhere to render all sensor readings.

jwise:
Fixes #493.


WolfwithSword:
Verified working: 
vncviewer_bFRTkUOQc2

WolfwithSword (approved):

PR #514, opened 2025-07-31
forward_shim: add eth0 to eth_list of interfaces to check in wait_for_network

jwise:
forward previously could not download a 3mf without WiFi up, even if netservice was happy about the eth0 situation, and so prints would hang on 'downloading'.  We help it out some.


jwise:
works for me and other users

PR #515, opened 2025-08-01
versions: upgrade to filament database 02.01.00.25 (adds PLA Tough+)

jwise:


jwise:
CI buil dtested and worked for me

PR #543, opened 2025-12-11
polar_cloud: Fix AMS data format and improve print handling

vmorrisPolar:
polar_cloud: Fix AMS data format and improve print handling

- Fix AMS status format to match Polar Cloud expectations: send nested ams array with humidity/temp per unit and tray arrays, instead of flat ams_tray list
- Add door field to status updates (always 0, X1 Carbon lacks door sensor feedback)
- Detect filament changing state: report status 10 when mc_print_stage == 1 while idle
- Fix print file handling: determine 3mf_file vs gcode_file action based on URL extension, not field name
- Ensure downloaded files have correct extension appended to filename
- Handle multiple field naming conventions from cloud (jobId/job_id, jobName/job_name, etc.)
- Extract ams_tray from print_options.ams_tray, ams.ams_tray, or direct ams_tray
- Fix bug: self.status = 0 instead of self.status == 0 (assignment vs comparison)
- Preserve AMS data in mqtt.py when push_status messages don't include it


jwise (approved):
LGTM

PR #547, opened 2026-01-06
AMS HT and AMS 2 Pro support

jwise:
Incomplete, but here is a pull request so folks can begin testing this.


L0g1calMadn3ss:
So this works? I just need to replace certain files on the SD card with those that you changed? What's this still waiting on to officially go through? 

I thought I was gonna have to have a local LLM teach me to hack the signals for reproducibility (CoPilot told me Bambu changed things and the original team gave up making AMS 2 work, or something). I also thought this would be a pretty big prio for more than just me, how the... am I the only other person to comment here?!?

No matter what, thank you for all this work! I finally hit my X1C's 1 year warranty and swapped to X1Plus without a second thought...........

teambvd:
> ... am I the only other person to comment here?!?

@L0g1calMadn3ss  I'd love nothing more than to be able to fully divorce myself from the closed source side of things. For me though, having ams2 + ht support was a big enough deal that I went back to stock. While having the printer effectively be a fully black-boxed device, one which I've simply firewalled/siloed off into 'I trust you less than <insert untrustworthy thing>' network vlan isn't exactly desirable, I don't necessarily see a ton of benefit to going back to X1Plus at current.

The granular speed control during printing would probably pull me back over, especially now that I'm trying out some of these 'foamy' flexibles out there with variable shore hardnesses, but other than that - well, the printer just kinda 'does it's job', and dutifully prints away everything thrown at it, till someday it'll die and, if I'm lucky, propel me to INDX 😁

L0g1calMadn3ss:
@teambvd Think I'll just buy a Snapmaker U1 😂

jwise:
rebased on top of main

jwise:
OK so:

1) the problem where I was having feed torque issues ever since I installed the AHB?  It is because I installed the blue and black coupler on the back of the printer backwards.  But,

2) I added a button to back down the MC firmware version just in case.

That was the last thing I wanted to do before merging this to main.  So I'm going to do that now.

PR #550, opened 2026-02-25
aiocamera, polar_cloud: improve camera failure handling

vmorrisPolar:
Raise RuntimeError in receive_h264() when the RTSP stream ends without delivering a complete frame (SPS + PPS + IDR), rather than returning None and causing a confusing TypeError in the caller.

Add a 15-second timeout to the receive_jpeg() call in the polar_cloud status loop, so a hung RTSP connection doesn't block status updates from being sent to Polar Cloud indefinitely.


jwise:
This looks correct to me but it has an errant commit in there (it has the commits that I already merged onto master).  I'll try to manually merge it sooner or later.  I was going to try to get an X1Plus 4.0 out with AMS2/AMSHT support -- and with these fixes -- but there were some issues with this and I kind of ran out of steam...

PR #554, opened 2026-04-14
ipcam Watchdog and Filament Remainder display implementation

mikefromdot:
Implementation of watchdog to address ipcam issue that causes camera to hang until powercycle
as well as
Implementation of screen patch which displays estimated filament remaining from ams in the filament tab.

 (fixes #553 and implements #532)

Issue #435, opened 2025-01-30
USB stick support to transfer Gcode files to printer on X1Plus Extension board


Durksz:
First of all thank you for hacking the Bambulab printers and making opensource firmware for it!

Problem statement:
The company I work for, insert multionational company name here, has a policy that only encripted usb sticks are allowed.
The fact we are using micro-sd cards to feed our bambulabs at the R&D department is tolerated, for now. 
And Plugging it in to the network (Wired or wireless) is a big nono (only for updates we hook it up wifi).

So having a usb input for feeding our models into the Bambu Lab would be very very usefull

Since the Expander board has 2 USB inputs, it would be ideal for the job.

I sincerely hope you can put this feature on the roadmap for implementation in the X1Plus firmware in the near future.

RobinWhitfield:
USB sticks encrypted with what exactly?

The printer would also need to be able to handle that, which could be more difficult.

bani6809:
> The company I work for, insert multionational company name here, has a policy that only encripted usb sticks are allowed.

The possibility of support depends entirely on what encryption is used. Can you be more specific about your encryption?

Durksz:
I will try to find out what encryption is used. I Will share brand and type of usb stick we received. Once i get back in office next week.

If my memory is correct it is that reading is always permitted but writing from laptop to usb / sd is not.

Durksz:
the USB stick that is provided to us is a Apricorn NX with FIPS140-2 Level 3 validated on it.
As far as i understand it is an hardware locked usb-stick (we need to type a code to unlock it).

To be honest, being able to use an unencrypted usb stick already can be usefull, since you cannot use the micro-SD card easily anymore to transfer printfiles to the printer. When the sd card has X1Plus firmware installed on it. We have 20 or more people using the same printer....

Durksz:
I'm no proficiant programmer (mechanical engineer with some limited python / arduino programming skills), is it complicated to implement a usb-reader feature?

jwise:
USB support would be nice, but I don't think it's going to make the initial Expander release.  Leaving the bug open to track the request, though.

mattcaron:
Digging a little bit, this appears to just be a UI issue.

Plugging in a USB stick (normal, FAT, not encrypted), it gets mounted properly automagically.

```
/dev/sda1                57.7G     34.1M     57.7G   0% /media/usb0
```

But, there's no way to tell the printer to print the file on it.

mattcaron:
[...]

PR #558, opened 2026-05-21
bbl_screen: show filament remaining percentage on FilamentPage

### LINKED ISSUE #532: Feature Request: Display Exact Filament Percentage for Each AMS Spool in X1Plus GUI
### **Description:**
In the current X1Plus GUI, the AMS spools are represented visually but without an exact numerical indication of the remaining filament.
While the graphical bar gives a rough visual idea, it lacks precision, making it harder to judge at a glance whether a given spool has enough material left for the next print.
Since Bambu printers already transmit the remaining filament data (in %) via MQTT, as confirmed by the Home Assistant Bambu Lab integration, this value is already available at firmware level and could be easily displayed in the GUI.

### **Suggested Enhancement**
Add a numerical percentage indicator (e.g., 69%, 23%, etc.) directly on or near each AMS spool representation in the X1Plus GUI.
Benefits:

- Provides an exact quantitative readout instead of a rough visual bar.
- Eliminates the need to check filament levels through external tools.

### **Reference Implementation**
The Home Assistant Bambu Lab integration already displays exact filament percentages for each AMS slot, using data exposed through MQTT.
This demonstrates that the value can be reliably obtained and visualized (see screenshots below).

Image  

Image

### **Visual Example (Mockup)**
Here’s a conceptual mockup of how the feature could look within the X1Plus GUI — the percentage is shown directly over each AMS spool icon:

Image

**Software Information**
 - Current X1Plus Version: 3.1

### **Additional Context**
This addition would make X1Plus even more self-sufficient as a local interface, reducing reliance on external integrations or slicers to access this crucial information.
It would also align the GUI’s usability with tools like Home Assistant which already expose this same data.
Related feature requests for reference:
https://github.com/SoftFever/OrcaSlicer/issues/9316
https://github.com/bigtreetech/PandaTouch/issues/352

mikefromdot:
Adds filament remaining percentage below the filament type label on the FilamentPage. The percentage is only shown when `modelData.remain` is non-negative (i.e. the printer has valid remaining data). The label is shifted up slightly to make room.

This was broken out as a focused PR from #554.

### DIFF
--- a/bbl_screen-patch/patches/printerui/qml/printer/FilamentPage.qml.patch
+++ b/bbl_screen-patch/patches/printerui/qml/printer/FilamentPage.qml.patch
@@ -17,7 +17,19 @@
      MarginPanel {
          id: howToPanel
          height: 538
-@@ -651,36 +652,49 @@
+@@ -633,8 +634,8 @@
+                     Text {
+                         width: parent.width - 10
+-                        y: 50
++                        y: 30
+                         anchors.horizontalCenter: parent.horizontalCenter
+                         font: Fonts.body_26
+                         minimumPixelSize: 10
+                         fontSizeMode: Text.HorizontalFit
+                         horizontalAlignment: Text.AlignHCenter
+@@ -650,37 +651,50 @@
+-                        text: modelData.exist ? (modelData.typeName + "").replace(" ", "\n")
++                        text: modelData.exist ? (modelData.typeName + "").replace(" ", "\n") + (modelData.remain >= 0 ? "\n" + modelData.remain + "%" : "")
                                                : qsTr("Empty")
                      }
  


mattcaron:
Tested this as part of the build provided in the comments in #435.

Works perfectly, that is, all previous behavior for X1Plus 3.1 is retained, there is just now a % which decreases as printing happens. I had a mix of spools in and the percentage was roughly correct. I assume that you're not doing math here - you're just using a number returned by the firmware in the AMS. If that is correct, then this is as good as it gets. If it is incorrect, then we might be able to get some better math, especially at lower levels (i.e. I had one that is maybe 5% full and it reads 0%.).

One question that I wanted to put to the group for discussion, however:

**Is there a point to having both a percentage and the existing "fuel gauge" style fill level?**

Personally, I don't like the fuel gauge type, because you can't see the color once it gets down too low. I'd rather there be a solid color box with a percentage all the time. Now, this doesn't matter when you're standing in front of the printer, but it matters (a bit) when using the remote display.

I have no strong opinion here, I just wanted to ask the question. As far as I am concerned, this is good to go as-is.

mikefromdot:
@mattcaron 
I did it to be honest because someone had asked for it in an issue. I found it to not be that accurate myself, it had a pretty wide margin of error. 
You are correct in that this is just pulling what the AMS/Printer is reporting so BBL's math is pretty bad.

jwise:
If it works for you, I'm ok with it

PR #559, opened 2026-05-21
bbl_screen: USB Implementation

mikefromdot:
Adds a USB storage file browser dialog to the printer UI, allowing users to browse and interact with files on USB storage devices.

Changes:
- `UsbStorageDialog.qml`: new dialog for browsing USB storage
- `interpose.cpp`: updated interpose to support USB storage integration
- `Expansion.qml`: updated expansion settings page

This was broken out as a focused PR from #554.


mattcaron:
Tested this as part of the build provided in the comments in #435 .

**Explicitly not tested:**
* Copying to and from the USB drive using the UI.
  * I swear I saw something about this, but couldn't figure out how to do it, so I didn't.
* Printing via the Expander Hardware drop town - now that the main UI supports it, it's pretty unergonomic to get there.
* Printing via USB via a USB hub. Because... why?

**Tested:**
* Installs & Boots - Yes
* Print from SD - Works as normal (no regressions)
* Print from USB0 via normal file menu - Works with caveats (see below)
* Print from USB1 via normal file menu - Works with caveats (see below)
  * 2 drives inserted, both with different files.

**Caveats:**
1. When printing from USB, it auto detects the wrong filament. I *think* it is always defaulting to A1, but am not 100% certain on this. Reproduction scenario (at least, this caused it to fail consistently for me):
  a. Install Grey PETG in A1
  b. Install Grey PLA in A3
  c. Go to print a model requiring Grey PLA
  d. The UI shows Grey PLA with a slot of "--" and then proceeds to use the PETG (which, of course, goes all spaghetti and fails)
    * **Workaround:** Manually choose the correct filament by tapping on the selection, then choosing it before hitting print.
2. When a print from a USB drive is completed, the results screen lacks the "print again" option present when printing from SD card.
3. USB directory listings can be stale. Reproduction scenario:
  a. Plug in a USB stick, print from it (note: just browsing it *might* be sufficient, a print may not be necessary).
  b. Once the print is complete, leave the screen open to that directory listing.
  c. Pull the USB stick out.
  d. Delete the files and replace them with different files.
  e. Plug the USB stick back in.
  f. It will show the old group of files, not the new ones.
    * **Note** I might have just been impatient and not waited the 5 second polling interval - but it should have at least cleared out the stale listing once it noticed that the drive went away, right?
    * **Workaround** Tap on the SD card tab, then tap back to USB.
4. Multiple plates don't work. Reproduction scenario:
  a. Create a `.gcode.3mf` with multiple plates.
  b. Put it on a USB stick
  c. Go to print from the USB stick
  d. You can only print the first plate.
    * **This is the only thing that I would consider a showstopper for release** as printing from USB is seriously less useful without the ability to batch up jobs like this (I really don't want to go back to exporting plates one at a time).

**Other notes:**
* I received the dreaded "system state is abnormal" warning. Rebooting didn't fix it - it returned pretty much as soon as I started printing. Prints worked fine, however.

**Questions:**
1. Large models show a "Gcode download" screen. Cool.
  a. Where is it downloading to? Internal? SD card?
  b. Is it deleted when the print is completed? Basically, are we creating a housekeeping nightmare where our SD card fills up?
10. Timelapse is still saved to the SD card. Is it possible to have it save to the USB stick from which the file is bring printed? This would save wear / filling up on the SD card. (This SHOULD NOT be addressed in this PR. Call it a followup feature request).

mikefromdot:
I think I fixed most of these actually for a separate PR which adds the usb to the normal files UI except for the multi plate thing (I dont print multiplate so I didnt think about it).
Does your USB ui show a system info file? If it does its the version where you just move files to the SD from the USB.

edit: I actually just looked, you actually have a build that is using changes that are not included in this PR. Thank You for the feedback on that, I'll look into those issues.
This PR is just for adding the USB to the expander menu and copying from it to the sd card

mattcaron:
Specifically talking about the "copy files" thing, I have USB in the expander menu, but:

1. Despite 2 drives being connected, I can't figure out how to select between them.
2. There is no sysinfo file. All see is the root USB device, the files on it, and options to format drive and close.

Now, that said, that may not be the code in *this* PR. But, like I said, I don't have a build environment set up right now. However, if you want me to test something specific, I'm happy to do so if you provide a build like you did before.

To be clear - I feel really bad that I don't have time to set up a build environment and hack on code with you, and I am super grateful that you have the time and inclination to do it. Unfortunately, the only help I can give the project right now is to test things.

mikefromdot:
ah ok I never put in logic to select a drive, thats probably why on that.
If you want we can start a thread in the discord for testing.

mattcaron:
Sorry, no. I ragequit Discord after they decided they hated freedom. Since they've never apologized, I refuse to return.

I really wish this project had a Matrix chat. It's actually decentralized free speech.

You could also just email me at the email in my profile.

mikefromdot:
Ok, I've fixed some things that @mattcaron noted for just the original implementation of USB and squashed them into one commit for review.

Matt I'll shoot you an updated test build with all of the changes I made as well as what branches they are coming from.

jwise:
I gave you push permissions so that you can push to a branch of your own (and so that CI will work properly).  Please prefix your branches with `mikefromdot/` when pushing.

I still haven't reviewed this yet but I at least reviewed and merged your other two PRs, and rebased ams2 support on top...

jwise:
At a skim --

I need to take a closer look at the actual logic inside UsbStorageDialog but it all seems pretty well contained so I'm relatively go "probably good enough!" at that one.

Expansion.qml has a lot of whitespace changes -- looks like line endings changed?  Can you put those back to unix line endings?

I'm ok with UsbStorageDialog being in an external qml file (and maybe all of our custom QML should be in an external QML file for ease of hacking...), but it looks like there are two copies of it.  Remove the old copy?

mikefromdot:
Danke for that, 
Changes made and set to be pushed as mikefromdot/usb

mikefromdot:
Pushed fixes to mikefromdot/usb per your feedback — fixed Expansion.qml line endings and removed the duplicate UsbStorageDialog.

bbl_screen-patch/patches/printerui/qml/settings/Expansion.qml
```diff
@@ -15,7 +15,27 @@ import "../printer"
...
+    property var usbDrives: []
+
+    function refreshUsbDrives() {
+        var result = X1PlusNative.popen("awk '$2 ~ /^\\/media\\/usb/ {print $2}' /proc/mounts");
+        if (!result || result.length === 0) {
+            usbDrives = [];
+        } else {
+            usbDrives = result.split("\n").filter(function(s) { return s.trim().length > 0; });
+        }
+    }
```
jwise: I wonder if we should control the mounts, rather than letting the system automounter deal with it -- especially if we are going to control multiple USB devices.

And if we do that, we can know what devices are plugged into which ports!  e.g., /sys/bus/usb/devices/1-1.3 will always be one of the USB ports, and 1-1.2 will always be the other one.

bbl_screen-patch/interpose.cpp
```diff
@@ -1,1065 +1,1085 @@
-/* bbl_screen interposer
```
jwise: I think something went wrong with the whitespace on this also

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
```diff
@@ -0,0 +1,292 @@
...
+import X1PlusNative 1.0
+import "qrc:/uibase/qml/widgets"
+import "qrc:/printerui/qml/dialog"
+import "qrc:/printerui/qml/X1Plus.js" as X1Plus
+
+Item {
+    property alias name: textConfirm.objectName
+    property var drives: []
+    property var selectedDriveIndex: 0
+    property var rootPath: drives.length > 0 ? drives[selectedDriveIndex] : ""
```
jwise: Hm, this `drives` doesn't get reset by the upstream `usbDrives`'s timer, so this screen could get out of date.

I think the pattern that we tend to use in X1Plus for this is to have the `X1Plus.js` subsystem poll on things with timers like this, and expose things as a property.  That way, the logic is centralized, and all downstream users (like this one) are in sync with each other.  (You coul dimagine that that `X1Plus.js` subproperty would also be responsible for, for instance, handling reformat state machines and such.)

bbl_screen-patch/interpose.cpp
```diff
@@ -1,1065 +1,1085 @@
...
+            QDir::DirsFirst | QDir::Name | QDir::IgnoreCase);
+        QJsonArray result;
+        for (const QFileInfo &info : entries) {
+            QJsonObject obj;
+            obj["name"] = info.fileName();
+            obj["isDir"] = info.isDir();
+            obj["size"] = (qint64)info.size();
+            result.append(obj);
+        }
+        return QJsonDocument(result).toJson(QJsonDocument::Compact);
```
jwise: I have no idea what the idiomatic way is to do this in Qt, but I'm pretty sure it's not passing JSON strings back and forth.  On the other hand, I can't come up with a better way, so ...

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
```diff
@@ -0,0 +1,292 @@
...
+        currentPath = path;
+        selectedEntry = null;
+        copyStatus = "";
+        try {
+            entries = JSON.parse(X1PlusNative.listDir(path)).filter(function(e) {
+                return e.name.charAt(0) !== '.' && e.name !== "System Volume Information";
+            });
+        } catch(e) {
+            entries = [];
+        }
```
jwise: DRY: call into `refreshEntries` directly (or have `entries` be a binding on `currentPath`, and then periodically set `currentPath = currentPath` in the timer or something??)

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
```diff
@@ -0,0 +1,292 @@
...
+                   formatState === "formatting"? qsTr("Formatting...") :
+                   formatState === "done"      ? qsTr("Format complete!") :
+                                                 qsTr("Format failed")
+            visible: drives.length > 0
+            keepDialog: true
+            onClicked: {
+                if (formatState === "idle") {
+                    formatState = "confirm";
+                } else if (formatState === "confirm") {
+                    var dev = X1PlusNative.popen("awk '$2 == \"" + rootPath + "\" {print $1}' /proc/mounts").trim();
```
jwise: definitely it would be good for `X1Plus.js` to know about all of this state on its own to avoid having to repeat this a bunch -- and maybe to centralize the format logic there

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
```diff
@@ -0,0 +1,292 @@
...
+                    if (!dev || dev.length === 0) {
+                        formatState = "failed";
+                        formatResetTimer.start();
+                        return;
+                    }
+                    formatState = "formatting";
+                    X1PlusNative.system("umount '" + rootPath + "'");
+                    var ret = X1PlusNative.system("mkfs.vfat -F 32 '" + dev + "'");
+                    X1PlusNative.system("mount '" + dev + "' '" + rootPath + "'");
+                    if (ret === 0) {
```
jwise: I am a little anxious about having format support at all.  For one, the way we format here does not necessarily clean up all kinds of bad formattings (a bad partition?).  And there are some weird things going on in Bambu's scripts (see `/usr/bin/format_sdcard.sh`) that this does not replicate.

And for another, it seems like an excessively destructive task that most users will not want to do from the printer.  That is to say, there are a lot of things that can go wrong, and the risk seems like it outweighs the usefulness to users.

Do you have a use case that formatting makes sense in that you're thinking about?  If not, I wonder if we should just get rid of formatting support.

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
```diff
@@ -0,0 +1,292 @@
...
+        DialogButtonItem {
+            name: "copy"
+            title: copyStatus !== "" ? copyStatus : qsTr("Copy to SD card")
+            visible: selectedEntry !== null && !selectedEntry.isDir &&
+                     selectedEntry.name.slice(-10).toLowerCase() === ".gcode.3mf"
+            keepDialog: true
+            onClicked: {
+                var src = currentPath + "/" + selectedEntry.name;
+                X1PlusNative.system("mkdir -p /sdcard/usb_imports");
+                var ret = X1PlusNative.system("cp '" + src + "' '/sdcard/usb_imports/'");
```
jwise: using `system` on unsanitized string interpolation makes me anxious (you could definitely get code exec with this), even if the input did nominally come from the user.  use `X1PlusNative.readFile` and `X1PlusNative.saveFile` instead

jwise:
I conceptually like this.  I think my implementation notes are that it would be good to centralize some of this logic into `X1Plus.js`, and we should maybe think twice about whether we want to expose destructive actions.  I also haven't, you know, tested it on my own printer yet!  But it seems definitely plausible.

Can you edit the PR to have the merge compare target of it be from `X1Plus:mikefromdot/usb`?  If you do, then it will be able to build in CI, also.

bbl_screen-patch/patches/printerui/qml/settings/Expansion.qml
mikefromdot: hmmm that might actually be beneficial, especially with the SD offload feature that I was going to submit after this was good (right now it just defaults the first drive that gets plugged in, having choice is better)

Question is do we want that done here or in the eventual submission for SD offload?

bbl_screen-patch/interpose.cpp
mikefromdot: fixed this, I have been doing a lot of the initial stuff on windows via vscode and WSL. 
Moved to a dedicated linux machine going forward so that should stop.

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
mikefromdot: good point, will fix

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
mikefromdot: My thought process was that a user could pick up any drive and just reformat it for use, You are right though format_sdcard.sh does do some weird shenannigans.

Kind of seems like something we might want to just not have as a user is going to have to have a properly formatted drive anyways to put files onto.

I'm going to go ahead and remove it, i guess if someone REALLY wants it we can reapproach later.

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
mikefromdot: fixed

bbl_screen-patch/patches/printerui/qml/settings/Expansion.qml
jwise: Good question.  Given that there is the plan of doing SD offload (which presumably needs `x1plusd` support), maybe all of the mount/unmount stuff should live in x1plusd dbus, and we should just bite the bullet and do USB detection (and mount/unmount APIs?) in x1plusd expansion (and plumb that through to x1plus.js to expose as a property) rather than having that polling take place in QML?

I think that if there's the plan of doing a fast follow, I guess I don't see any particular need to do it the 'wrong' way first just to land something.

images/cfw/opt/x1plus/share/qml/UsbStorageDialog.qml
mattcaron: My vote is to remove it.

1. Everyone can format stuff as FAT32.
2. Most drives come that way to begin with.
3. So it's just a big fat "erase everything" button waiting to be fat fingered.

mikefromdot: iirc I did at least have it so you had to press it twice.

bbl_screen-patch/patches/printerui/qml/settings/Expansion.qml
mikefromdot: yeah, sd offload is 100% in x1plusd, its bundled with that QoL change I have going on my fork and that I shared the pic of. 

It probably does make sense to do it now rather than later tbh.

mikefromdot:
Alright. 
This should be cleaned up enough to review again.

Made changes to address feedback and updated to use `X1Plus:mikefromdot/usb`

edit: Learning how to properly git is "fun". It should really be fixed now.

mikefromdot:
@jwise 
I think I have mostly addressed your concerns with this one. You want to give it another once over?

jwise:
I think *something* went wrong here but I'm not sure exactly what.  I think this PR is set to have the X1Plus branch as the base, and your branch as the compare source (should be X1Plus:main as the base, X1Plus:mikefromdot/usb as a compare source).  But dang, if this is your first time using `git` in a production project, not bad!

Skimming the diff, it looks also somehow like X1Plus wifi driver config options were removed?

I think I'm going to have a maintenance day tomorrow, so I'll try to get to this tomorrow in more detail.

mikefromdot:
Yeah I think its being ornery because the PR was originally for main.

The WiFi driver:
Ahhh that. I think I know why and thats my bad on the fork side.
I'll fix that, it shouldnt have been in here but it is. Thats supposed to be in another branch on my end (QoL changes) but I think it got a little comingled.

mikefromdot:
X1Plus Menu
This is what is planned to be submitted after, that will live in the X1Plus tab.
I felt that having all the x1plus specific stuff have its own menu was a little easier for end users

jwise:
Now that I finally got #547 landed, I rebased this onto main in the branch `jwise/usb`.  There are probably a few things I want to change (move the `usbInfo` thing into x1plusd also?) but I like the `UsbDrivesChanged` / `GetUsbDrives` thing.  I also want to play with the UI a little.

mattcaron:
I've been quiet, but I've been testing a build @mikefromdot sent me for about the past week. I have about 3 pages of notes to type out here, I just haven't had the time. Hopefully I'll find it by mid week. I would strongly advise against merging this until you at least read through my testing results, because there's some pretty odd stuff in there (phantom USB devices, increasingly laggy UI that feels like a memory leak, etc.). 

Overall, the beginnings of this are really good, it just needs some testing and polish.

mikefromdot:
Yeah the build you have was everything all together at the time.

This one was just the USB support base and Josh has pulled it into his own branch and is polishing it up further.

jwise:
I just tried this out on my printer.  It appears to work.  I'm working on changing the manager class to encapsulate the port mappings in each Expansion's device definition class, and having it poll USB devices directly (to give vid/pid data for unsupported devices) and drives with storage capacity directly, and assign nicer names to each (especially when on the other side of a hub).  I started in on this, though I think I'm going to try to go get some coffee and maybe go climbing for a bit.

mattcaron:
> I just tried this out on my printer. It appears to work. I'm working on changing the manager class to encapsulate the port mappings in each Expansion's device definition class, and having it poll USB devices directly (to give vid/pid data for unsupported devices) and drives with storage capacity directly, and assign nicer names to each (especially when on the other side of a hub). I started in on this, though I think I'm going to try to go get some coffee and maybe go climbing for a bit.

Go climbing. I finished mowing the lawn, tilled the corn field, and am now pulling some Cat6 so my wife's Peloton will stop dropping the WiFi like a hippie in the 60's dropped acid.

But... if I can get that finished, I might have time to type up my notes. I've basically been running it constantly, printing OpFor for the Forgotten Ruin Adventure games (you need 3 dozen of 20 different types of fantasy bad guys for your modern soldiers to fight).

So, I might have input by the time you get back.

mattcaron:
Okay, I found some time before dinner...

Testing `3.1-USB2.x1p` that @mikefromdot sent me (which hopefully will be the last private, untracked build because I should just be able to grab CI builds from now on..)

* Boots? **Yes**
* Print from USB0 works? **Yes**
* Print from USB1 works? **Yes**
* Print from SD works? **Yes**
* Copy files from USB works? **Yes** (I found out how to do it!)
* `Print Again` works when printing from USB? **Yes** (but see notes)
* Choosing which plate to print off a multi plate file works when printing from USB? **Yes** (but see notes)
* USB refresh works (unplug, add files, replug)? **Yes** (after expected 5 second polling delay)

**Notes:**

1. `Print Again` works, but it reprints exactly the same plate you just printed. When printing from SD, it lets you choose the plate you want to print. So, it's not "print the same plate", it's "print the same file" (which lets you pick the plate.
    * This is minor and can be worked around, but it is a difference in behavior that people will notice.
2. When selecting the plate to print via USB, the arrows are always on both sides. When selecting the plate to print via SD card, the first plate only shows a right arrow, the last only shows a left arrow, and if there is only one plate, it shows no arrows.
    * Again, minor.
3. SD card offload defaults to "On"
    a. Is this behavior desirable?
    b. Should we rename it to something that references USB? Perhaps "Offload to USB"? Not sure which is more clear.

**Old bugs**
1. Printing from USB still doesn't pick the right filament.
    a. It doesn't do the "--" anymore, but it always seems to default to the first slot with filament in it.
    b. Example: I wanted to print something in silver PLA. I had silver PLA in A2. It defaulted to the green PLA in A1.
2. It still reports "System state abnormal".

**New bugs**
1. When I boot with 2 USB drives connected, "extra" drives show up.
    a. The first time, I had drives USB0, USB1, and USB2. Only 0 and 2 were real drives with files.
    b. The second time, I had drives USB0, USB2, USB3, and USB4. 4 was a dupe of 0 (both showed the same files), 3 was real, and 2 was fake. There was no 1.
    c. Note that the behavior with 0 and 1 USB drives connected is expected (0 or 1 drive shown).
    d. I realize that @jwise is reworking how the USB enumeration works, so this is likely worth keeping an eye on while doing that rework.
3. After running and printing constantly from USB for about 3 days (call it plates), the UI became so sluggish as to be unusable. As in, you tap "home" and it takes 30 seconds to change the screen.
    a. Power cycle fixes it.
    b. Repeated twice.
    c. Feels like a memory leak.
4. I got a transient "Unable to record timelapse without MicroSD card inserted" error, but it disappeared without any actions on my part. It never showed up in the messages pane. It only showed up in the side panel of messages while printing and, as I said, disappeared within about 10 seconds of the screen waking up.
    a. "SD card offload" was enabled.
    b. I had 2 USB drives inserted and was printing from USB1.

I think that's all. When there's a new build to test, I'll grab it, reflash, and start testing with that.

Thank you both for all your hard work!

mattcaron:
So, update - long term testing results.

I have basically been printing constantly for the past 6 weeks (so many boardgame components... Aliens: AGDITC, Fate of the Nostromo, Mister Torgues, etc...)

I have not found any bugs except as already noted above.

However, the "It won't pick the correct filament" bug is really amorphous. Sometimes, it works great and picks the right filaments. Other times, it picks the wrong ones. I have no idea why. There seems to be no pattern to it.

PR #560, opened 2026-06-04
fix liveview error from camera_watchdog.py

mikefromdot:
change camera_watchdog to allow for polling of the service instead of just assuming that 8 seconds is enough time.

This should prevent the liveview error that has been popping up


images/cfw/opt/x1plus/lib/python/x1plus/services/x1plusd/modules/camera_watchdog.py
```diff
@@ -195,11 +195,12 @@ async def _stop_ipcam(self, rtsp_enabled: bool):
...
-            logger.info(
-                f"waiting {RESTART_SETTLE_TIME}s for ipcam to settle "
-                "before restoring RTSP state"
-            )
+            logger.info("waiting for ipcam to become ready before restoring RTSP state")
             await asyncio.sleep(RESTART_SETTLE_TIME)
+            while not await self._probe_camera():
+                logger.debug(f"ipcam not yet ready, retrying in {BOOT_POLL_INTERVAL}s")
+                await asyncio.sleep(BOOT_POLL_INTERVAL)
+            logger.info("ipcam ready, restoring RTSP state")
```
jwise, line 203: This won't work at all, will it?  I thought that `_probe_camera` won't pass until RTSP is already awake.

mikefromdot:
So the original just fired it off about 8 seconds after ipcam was told to
start and toggled the button for rtsp causing the error.

This actively waits until it can reach it before it toggles

jwise:
Yes, but it won't be able to reach it until it is first toggled, right?  `_probe_camera` requires RTSP to be turned on in order to pass, I thought?

mikefromdot:
So it does a TLS handshake to see if ipcam is up which doesnt rely on rtsp being enabled. When ipcam dies it just stops ack'ing the handshake which fires off the watchdog. 

It seems weird to me too but I figure ipcam is always running regardless of whether you have liveview on or not.

The way watchdog used to handle reenabling rtsp was just to wait 8 seconds after starting ipcam and just toggling the rtsp if it was enabled, this change will make it wait until it gets ipcam to ack before it toggles.

RebeccaStevens:
I tried this PR locally, but unfortunately it didn't fix things for me.
(No idea if you wanted this feedback or not, but I thought I'd give it in case).

mikefromdot:
hmmm, I havent had a liveview error with it myself with this but I could be the outlier with it.
If you could grab logs from around when the cam locks up and throws the error that would be great.

RebeccaStevens:
Sorry for not getting back to you sooner.
Just to confirm that we're talking about the same thing. What I'm experiencing is that with the Live Camera Feed (in both OrcaSlicer and BambuStudio), when I click play, it will play for about 10-30 seconds and then freeze. I have to close and reopen the slicer to be able to play again - the feed will break again after 10-30 seconds.

Which particular logs would like? Just point me at the file and I'll attach it.

